PR #55, opened 2018-11-19
[WIP] Model tests

kegl:


kegl:
I finished covering model and db_tools. Got out all commits from model.py. I'd prefer to merge now and get to views.py once we have a working version with all this. If the local training engine works then we can also eliminate a lot of code from db_tools.

@aboucaud @jorisvandenbossche @glemaitre

PR #70, opened 2018-11-23
ENH: AWS Worker independent of database

jorisvandenbossche:


jorisvandenbossche:
@kegl @aboucaud @glemaitre this should be ready for review.

It's working locally, with code like this:

```
conf = read_backend_config(os.path.join(HERE, 'config.yml'))
worker = AWSWorker(conf['aws'], submission='local_starting_kit',
                       ramp_kit_dir=ramp_kit_dir)
worker.setup()
worker.launch_submission()
worker.collect_results()
worker.teardown()
```

which trains the submission and downloads the logs and predictions from amazon.

I didn't test the `train_loop` which is currently being used on the server (I updated it for the changes I made, but for sure there is something I will have missed). It would be nice to have a way to test this, but on the long term I think we also want to replace this code with a generic loop that uses the worker interface.

ramp-engine/rampbkd/aws/worker.py
```diff
@@ -0,0 +1,76 @@
+import logging
+import os
+import sys
+
+from ..base import BaseWorker
+from . import api as aws
+
+
+logger = logging.getLogger('ramp_aws_engine')
+logger.setLevel(logging.DEBUG)
```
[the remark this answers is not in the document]
glemaitre: I would say that we should not set it here. It should be the user that start a dispatcher or a worker to set the level and where to log (file or stdout)

glemaitre:
Some remaining waste from the merging.

```
ImportError while importing test module '/home/travis/build/paris-saclay-cds/ramp-board/ramp-engine/ramp_engine/tests/test_aws.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
ramp-engine/ramp_engine/tests/test_aws.py:16: in <module>
    from rampbkd.aws import AWSWorker
E   ModuleNotFoundError: No module named 'rampbkd'
```

jorisvandenbossche:
I tested it locally and the tests (both just the worker, as integrated with the dispatcher) work. 

There is only still a problem that, when used from the dispatcher or ramp-test, there are failures in connecting to AWS. Not yet fully sure why that is, but we might need to add some more retries or give certain actions more time.

doc/workers.rst
```diff
@@ -19,3 +19,93 @@ Available workers:
...
+specific resources (e.g. GPU's).
+
+You need to create an Amazon Machine Image (AMI) with the starting kit and
+data, and the needed packages to run submissions. This AMI will then be used
+to launch instances to run submissions.
+
+A very short how-to for creating such an AMI manually:
+
+- launch an Amazon instance
+- connect to it
```
glemaitre: remove trailing space

ramp-engine/ramp_engine/aws/ramp_aws_train.py
```diff
@@ -5,12 +5,14 @@
 import argparse
 from argparse import RawTextHelpFormatter
 
-from ramp_engine.aws.api import launch_ec2_instance_and_train
-from ramp_engine.aws.api import train_on_existing_ec2_instance
 from ramp_engine.aws.api import validate_config
 from ramp_engine.config import read_backend_config
 from ramp_database.tools import get_submission_by_name
 
+from .aws_train import (
```
glemaitre: Do you want ot put it on 2 lines

ramp-engine/ramp_engine/aws/worker.py
```diff
@@ -0,0 +1,106 @@
...
+from ..base import BaseWorker
+from . import api as aws
+
+
+logger = logging.getLogger('RAMP-AWS')
+
+
+class AWSWorker(BaseWorker):
+    """
+    Run RAMP submissions on Amazon.
```
glemaitre: You might want to copy the docstring of the other worker here.

PR #71, opened 2018-11-23
MAINT: use environment.yml to create CI environment

glemaitre:


jorisvandenbossche:
This is a good improvement to clean up the environment used to test the full thing, but what I additionally meant in my other comment: we can make another environment (which doesn't need that much packages, only rampwf and the dependencies needed for the iris kit), that is then actually used for the CondaEnvWorker

glemaitre:
Oh sorry I did not read fully the comment. It makes a lot of sense.

PR #78, opened 2018-11-27
[MRG] MAINT: remove deprecation warning from wtforms

glemaitre:
Remove couple of deprecation warning due to WTForms


glemaitre:
@kegl @jorisvandenbossche @aboucaud This should be safe to merge

PR #80, opened 2018-11-28
[MRG] MAINT: moving databoard module one level down

aboucaud:


glemaitre:
Can we do that after my PR :)

aboucaud:
Then hurry up, I'm almost green :)

jorisvandenbossche:
Can you try merging this in your PR? (just as a try, we were wondering if that would work :-))

jorisvandenbossche:
But something else, I would try to finish your current PR, and do a futher batch of clean-ups in a next PR. As there are already some things in your PR I would want to use for the testing.

glemaitre:
> But something else, I would try to finish your current PR, and do a futher batch of clean-ups in a next PR. As there are already some things in your PR I would want to use for the testing.

We can merge my PR now. I will just freeze what I have now. I would like that we fix the issue with the password checking thought.

aboucaud:
It's green now @glemaitre..just sayin'

PR #87, opened 2018-11-28
[WIP] REFACTOR: continutation of adding test, enhancement, and documentation

glemaitre:


glemaitre:
I need this PR to make the training loop which could use the worker.
I let some TODO which we can come in while to refine some of the part.

PR #93, opened 2018-12-17
EHN: Create dispatcher for RAMP workers

glemaitre:


glemaitre:
@kegl @agramfort @jorisvandenbossche You can have a first look at what we are aiming with the dispatcher and worker. It works locally (you need to fetch the branch that I opened in `ramp-workflow`).

We will have to loop with @jorisvandenbossche to decide how to handle the configuration which are passed.

glemaitre:
@aboucaud I think that I am almost done with the dispatcher.
Now, that I wrote mainly test for some of the `db_tools` from `databoard`, I actually see that you rewrote some of those in `rampdb.tools`. At least, I can move some of the missing tools there and try to merge with what I updated and port all the tests that I wrote. On the other side, I see that you recreate a `db` instance in each of the function. Is there any reason for that? I would assume that this is cleaner but it required to pass a config file each time, isn't it?

glemaitre:
I will merge this PR now (once travis is green). It is still experimental but it will allow me to start migrating the function required in `ramp-database` and add a serie of units test there.

aboucaud:
> On the other side, I see that you recreate a `db` instance in each of the function. Is there any reason for that? I would assume that this is cleaner but it required to pass a config file each time, isn't it?

No specific reason except I wanted to avoid keeping the connection alive and I did not take the time to implement a proper wrapper to the db context manager.

aboucaud:
Nice ! The dispatcher looks really neat.

glemaitre:
> No specific reason except I wanted to avoid keeping the connection alive and I did not take the time to implement a proper wrapper to the db context manager.

OK I see. So this is the clean version of some of the `db_tools`. Great, I see what I have to do now to migrate some of the things that I tested there.

@aboucaud I have another quick question. In `rampdb.tools.tools`, there is some `get_*` function but they are taking no session. I don't see how they can work without a connection and they are used nowhere. The same tools are in `db_tools` to call using `db`. So I am wondering why those tools are not interface in `query` and `api` like the other functions?

aboucaud:
These methods were defined within the db definitions in `model.py`, which is the only reason I put them here. Now I think that @kegl either replaced or deprecated them. If you don't find any occurence of them being used then you can get rid of them.

PR #101, opened 2018-12-21
[WIP] EHN: create ramp-utils to share tools across RAMP bundle

glemaitre:
This PR intend to create a package to pack tools shared across all the different RAMP package:

* management of the `config` file: we are going to have a single `config.yml` file and it seems that it is not the role of any of the package to handle part of the handling of this config.
* some module reuse some code which was duplicated: e.g., import from source


glemaitre:
This PR is temporary since there is no test in ramp database. I am going to introduce a dummy test just to be sure that travis stop to turn red as a false alarm. I will merge this PR then to be able to move to another PR on rampdb

PR #108, opened 2018-12-23
[WIP] MAINT: migrate db_tools and testing into ramp-database

### LINKED ISSUE #92: RAMP: Unit and integration tests
We need to have much more tests for all the different packages.

TODO:

- [ ] `ramp-database`:
    - [ ] tools
    - [ ] models
- [ ] `databoard`

### LINKED ISSUE #107: DATABASE: refactor rampdb.tools
The current code organisation in `rampdb.tools` is suboptimal. While it is interested to have `query` to simplify the code, I think that we should split the `api` into specific packages. The `tools` module is also a bit mysterious for the moment.

### LINKED ISSUE #109: DATABASE: session scope and connection to database
Both current way to connect to the database if wrong:

*  `databoard` is linked to Flask and the connection to the database is globally open for everyone and anytime.
* `ramp-database` open a connection and a session at each call of the session`.

As explained in the SQLAlchemy documentation, the session should be open when running a program and pass the session to each function avoiding to reopen a session each time. However, Flask is a bit too much permissive there. 

So all `get_**`, `make_***`, and `set_***` function should actually get a session object.
This object should be created at the scope of the application and more preferably with a context manager.

glemaitre:
This PR intends to move the following functionality into `ramp-database`:

- [x] some of the `db_tools` which are now tested in `databoard`;
- [x] the `testing` module which allows to set up some toy database for testing;
- [x] check how the configuration is used in the model: some currently use some call to some environment variable which is not correct. It should be passed when constructing the database;
- [x] implement/test update_leadearboad functions; moved to issue #110 
- [x] be sure that we send email (currently commented; moved to issue #111
- [x] port the test of `rampdb.tools.event` from `db_tools`.
- [x] Handle the bagged scores which is not managed for the moment; should be solved in: https://github.com/paris-saclay-cds/ramp-workflow/issues/167.
- [x] Create a deploy tool into `ramp-utils`.
- [x] Add documentation for `rampdb.model`
- [x] Add unit tests for the `rampdb.model`
- [x] Resplit tests in `rampdb.model` for easier failure debugging.
- [x] Document and test the backreferencing for the different tables in `rampdb.model`
- [x] Make the command from `tasks.py` accessible directly from a `cli.py`.
- [ ] All database created for the test should not be created in `/tmp/database_test' but instead, we should overwrite the configuration file and create a real temporary folder.

# reference issues:
closes #109 
closes #107 
closes  #92


glemaitre:
@kegl @agramfort @jorisvandenbossche 

So my migration is almost there regarding all the tools needed by the backend. The coverage is not bad (I still need to test the module `rampdb.model`).

However, by solving the last issue (i.e., removing the use of the environment variables in the database structure), I finally broke the structure of the database. The 2 places are:

* https://github.com/paris-saclay-cds/ramp-board/pull/108/files#diff-96a8b37606d895e2e57896339f4692c6R106
* https://github.com/paris-saclay-cds/ramp-board/pull/108/files#diff-2e23ed379aff8d52538e0c780900c309R39

The previous design (i.e. use of environment variable) is wrong. It does not allow to have different RAMP configuration at the same time. Having a configuration file is much better but we need to forward the information which is easily done using the `generate_ramp_config` which create all paths automatically which need to be forwarded.

So right now, I am almost sure that the backend is running and tested. I did not look at the front-end yet but I broke it for sure since my original test in `databoard` are not passsing.

So my questions are the following:

* What do you think about the changes?
* Are you all willing to merge it even if we break the front-end at the cost to repear it after. I can say that the solution to repear the front-end should be manageable. It is only passing couple of parameters in addition of the previous one.

glemaitre:
So deploying a RAMP event should be done with the following code:

```python

# read the example configuration for the iris event
config_file = path_config_example()

# This part is what the user needs to do to deploy the event
config = read_config(config_file)
ramp_config = generate_ramp_config(config_file)
deploy_ramp_event(config_file)

# simulate that we add users and sign-up for the event and that they
# submitted the starting kit
with session_scope(config['sqlalchemy']) as session:
    add_users(session)
    sign_up_team(session, ramp_config['event_name'], 'test_user',
                 ramp_config['ramp_sandbox_dir'],
                 ramp_config['ramp_submissions_dir'])
    submit_starting_kits(session, ramp_config['event_name'], 'test_user',
                         ramp_config['ramp_kit_submissions_dir'],
                         ramp_config['ramp_submissions_dir'],
                         ramp_config['sandbox_name'])

# ]This part is what the user needs to do after deploying the event
dispatcher = Dispatcher(config=config,
                        worker=CondaEnvWorker, n_worker=-1,
                        hunger_policy='exit')
dispatcher.launch()
```

Note that the middle part is actually the job of the front-end.

glemaitre:
Once the PR in ramp-workflow is merged, the test for the dispatcher will work.

glemaitre:
Ready to be reviewed. I would like this to be merged before to focus on the frontend. I might need @kegl for starting there.

ramp-database/rampdb/model/event.py
```diff
@@ -76,110 +159,265 @@ class Event(Model):
...
+            The session used to make the query.
+
+        Returns
+        -------
+        event_type_score : :class:`rampdb.model.EventTypeScore`
+            The default type score for the current event.
+        """
+        return (session.query(EventScoreType)
+                       .filter(EventScoreType.event == self)
+                       .filter(EventScoreType.name == self.official_score_name)
```
jorisvandenbossche: here you cannot use the `filter_by` as above?

ramp-database/rampdb/model/event.py
```diff
@@ -76,110 +159,265 @@ class Event(Model):
...
+                       .one())
 
     @property
+    # XXX: This will only work with Flask
     def official_score_function(self):
+        """callable: The default function used for scoring in the event."""
         return self.official_score_type.score_function
 
+    def get_official_score_function(self, session):
```
jorisvandenbossche: Are those additional functions actually used? (outside of the tests)

It is rather duplicative, and would be nice if we could avoid that

ramp-database/rampdb/model/event.py
```diff
@@ -76,110 +159,265 @@ class Event(Model):
...
     @property
     def n_participants(self):
+        """int: The number of participants to the event."""
         return len(self.event_teams)
 
 
 class EventScoreType(Model):
+    """EventScoreType table.
+
+    This is a many-to-many relationship between Event and ScoreType.
```
jorisvandenbossche: many-to-one?

If I recall the discussion, the ScoreType table was not that useful, and it is actually this table that stores the score types per event? (as one ScoreType record is created per record here) If so, I would mention something about that.

ramp-database/rampdb/model/fold.py
```diff
@@ -17,9 +17,26 @@
 
 
 class CVFold(Model):
-    """Storing train and test folds, more precisely: train and test indices.
+    """CVFold table.
 
-    Created when the ramp event is set up.
+    Storing train and test folds, more precisely: train and test indices.
```
jorisvandenbossche: This description is not fully clear to me.

So if I understand correctly: this table stores the train and test indices for a *single* fold ? And then each event has a bunch or records in this table, corresponding to the number of folds in the CV of that problem?

ramp-database/rampdb/model/problem.py
```diff
@@ -32,87 +25,178 @@
...
+
+    # XXX: big change in the database
+    path_ramp_kits = Column(String, nullable=False, unique=False)
+    path_ramp_data = Column(String, nullable=False, unique=False)
 
-    def __init__(self, name):
+    def __init__(self, name, path_ramp_kits, path_ramp_data, session=None):
         self.name = name
-        self.reset()
-        # to check if the module and all required fields are there
```
jorisvandenbossche: I think this can be useful to keep (so that you already get a complaint on init time when there is something wrong with your Problem)

ramp-database/rampdb/model/problem.py
```diff
@@ -32,87 +25,178 @@
...
-            encode_string(self.name), self.workflow)
-
-    def reset(self):
-        self.workflow = Workflow.query.filter_by(
-            name=type(self.module.workflow).__name__).one()
+        return 'Problem({})\n{}'.format(encode_string(self.name),
+                                        self.workflow)
+
+    def reset(self, session):
+        """Reset the workflow."""
```
jorisvandenbossche: it's not fully clear to me why it is needed or what it is doing?

ramp-database/rampdb/model/problem.py
```diff
@@ -32,87 +25,178 @@
...
 
     @property
     def Predictions(self):
+        """:class:`rampwf.prediction_types`: The predictions used for the
+        problem."""
         return self.module.Predictions
 
     def get_train_data(self):
-        path = os.path.join(RAMP_DATA_PATH, self.name)
+        """ndarray: The training data."""
```
jorisvandenbossche: This not necessarily a numpy array?

ramp-database/rampdb/model/problem.py
```diff
@@ -32,87 +25,178 @@
...
         return self.module.get_train_data(path=path)
 
     def get_test_data(self):
-        path = os.path.join(RAMP_DATA_PATH, self.name)
+        """ndarray: The testing data."""
+        path = os.path.join(self.path_ramp_data, self.name)
         return self.module.get_test_data(path=path)
 
     def ground_truths_train(self):
+        """ndarray: The true labels for the training."""
```
jorisvandenbossche: The return value here is not a ndarray, but a Predictions instance (and the same fot the two below)

ramp-database/rampdb/model/submission.py
```diff
@@ -56,16 +56,101 @@
...
+    """Submission table.
+
+    Parameters
+    ----------
+    name : str
+        The submission name.
+    event_name : str
+        The event name.
+    path_ramp_submissions : str
+        Path to the deployment RAMP submissions directory. It will corresponds
```
jorisvandenbossche:
```suggestion
        Path to the deployment RAMP submissions directory. It will correspond
```

ramp-database/rampdb/model/submission.py
```diff
@@ -56,16 +56,101 @@
...
+        to the key `ramp_submissions_dir` of the dictionary created with
+        :func:`ramputils.generate_ramp_config`.
+    is_sandbox : bool
+        Whether it is a sandbox submission.
```
jorisvandenbossche: Give a bit more details about what is a sandbox submission?

ramp-database/rampdb/model/submission.py
```diff
@@ -114,8 +206,14 @@ def __init__(self, name, event_team):
         # url which is maybe not a good idea.
         self.hash_ = '{}'.format(sha_hasher.hexdigest())
         self.submission_timestamp = datetime.datetime.utcnow()
-        event_score_types = EventScoreType.query.filter_by(
-            event=event_team.event)
+        if session is None:
```
jorisvandenbossche: a little bit above, `self.session` is already defined by inspecting the `event_team` object

ramp-database/rampdb/model/submission.py
```diff
@@ -56,16 +56,101 @@
...
+    training_timestamp : datetime
+        The date and time when the training finished.
+    contributivity : float
+        The contributivity of the submission.
+    historical_contributivity : float
+        The historical contributivity.
+    type : {'live' or 'test'}
+        The type of submission.
+    state : str
+        The state of the submission.
```
jorisvandenbossche: do we have a list of possible states somewhere?

ramp-database/rampdb/model/submission.py
```diff
@@ -56,16 +56,101 @@
...
+        The standard deviation of the computation time for a fold on the train
+        data.
+    valid_time_cv_std : float
+        The standard deviation of the computation time for a fold on the valid
+        data.
+    test_time_cv_std : float
+        The standard deviation of the computation time for a fold on the test
+        data.
+    max_ram : float
+        The maximum amount of RAM consume during training.
```
jorisvandenbossche:
```suggestion
        The maximum amount of RAM consumed during training.
```

ramp-database/rampdb/model/submission.py
```diff
@@ -56,16 +56,101 @@
...
 class Submission(Model):
-    """An abstract (untrained) submission."""
-
+    """Submission table.
+
+    Parameters
+    ----------
+    name : str
+        The submission name.
+    event_name : str
```
jorisvandenbossche: the init uses `event_team` ? And maybe it's not a string but an actual EventTeam object?

ramp-database/rampdb/model/submission.py
```diff
@@ -102,9 +187,16 @@ class Submission(Model):
     # later also ramp_id
     UniqueConstraint(event_team_id, name, name='ts_constraint')
 
-    def __init__(self, name, event_team):
+    # XXX: big change in the database
+    is_sandbox = Column(Boolean, default=False)
```
jorisvandenbossche: Is this needed? 
Meaning: is it needed that this name is configurable? We can also hard-code that any submission named 'starting_kit' is a sandbox submission?

ramp-database/rampdb/model/submission.py
```diff
@@ -102,9 +187,16 @@ class Submission(Model):
...
+    path_ramp_submissions = Column(String, nullable=False, unique=False)
```
jorisvandenbossche: It might be more logical to put this new column in the Event table? 
(I don't see a reason to allow it be differ between submissions for one event. I assume the reason to add it to the database in the first place is to allow to have a different directory for submissions *between* events)

glemaitre:
closing in favor of #112

PR #116, opened 2019-01-11
DOC: document the database model

jorisvandenbossche:
The docstring changes of @glemaitre splitted from https://github.com/paris-saclay-cds/ramp-board/pull/108


jorisvandenbossche:
@glemaitre first extracted all docstring changes in `/rampdb/model`, and in a second commit removed those things that I think are related to the additions to the database model you did in the other PR (which should not be here). Can you check if I removed the corrects things / all things ?

ramp-database/rampdb/model/event.py
```diff
@@ -89,82 +170,126 @@ def __repr__(self):
...
     @property
     def n_participants(self):
+        """int: The number of participants to the event."""
         return len(self.event_teams)
 
 
 class EventScoreType(Model):
+    """EventScoreType table.
+
+    This is a many-to-many relationship between Event and ScoreType.
```
jorisvandenbossche: many-to-one?

If I recall the discussion, the ScoreType table was not that useful, and it is actually this table that stores the score types per event? (as one ScoreType record is created per record here) If so, I would mention something about that.

ramp-database/rampdb/model/event.py
```diff
@@ -25,6 +25,87 @@
...
+    This table contains all information of a RAMP event.
+
+    Parameters
+    ----------
+    problem_name : str
+        The name of the problem.
+    name : str
+        The name of the event.
+    event_title : str
+        The title to give for the event.
```
jorisvandenbossche: used in the frontend? this one can contains spaces and the `name` not?

ramp-database/rampdb/model/fold.py
```diff
@@ -17,9 +17,26 @@
 
 
 class CVFold(Model):
-    """Storing train and test folds, more precisely: train and test indices.
+    """CVFold table.
 
-    Created when the ramp event is set up.
+    Storing train and test folds, more precisely: train and test indices.
```
jorisvandenbossche: This description is not fully clear to me.

So if I understand correctly: this table stores the train and test indices for a *single* fold ? And then each event has a bunch or records in this table, corresponding to the number of folds in the CV of that problem?

ramp-database/rampdb/model/problem.py
```diff
@@ -51,54 +73,82 @@ def __init__(self, name):
...
         return self.module.problem_title
 
     @property
     def Predictions(self):
+        """:class:`rampwf.prediction_types`: The predictions used for the
+        problem."""
         return self.module.Predictions
 
     def get_train_data(self):
+        """ndarray: The training data."""
```
jorisvandenbossche: This not necessarily a numpy array?

ramp-database/rampdb/model/problem.py
```diff
@@ -51,54 +73,82 @@ def __init__(self, name):
...
         path = os.path.join(RAMP_DATA_PATH, self.name)
         return self.module.get_train_data(path=path)
 
     def get_test_data(self):
+        """ndarray: The testing data."""
         path = os.path.join(RAMP_DATA_PATH, self.name)
         return self.module.get_test_data(path=path)
 
     def ground_truths_train(self):
+        """ndarray: The true labels for the training."""
```
jorisvandenbossche: The return value here is not a ndarray, but a Predictions instance (and the same fot the two below)

ramp-database/rampdb/model/submission.py
```diff
@@ -56,8 +56,82 @@
...
+    training_timestamp : datetime
+        The date and time when the training finished.
+    contributivity : float
+        The contributivity of the submission.
+    historical_contributivity : float
+        The historical contributivity.
+    type : {'live' or 'test'}
+        The type of submission.
+    state : str
+        The state of the submission.
```
jorisvandenbossche: do we have a list of possible states somewhere?

ramp-database/rampdb/model/submission.py
```diff
@@ -56,8 +56,82 @@
...
+        The standard deviation of the computation time for a fold on the train
+        data.
+    valid_time_cv_std : float
+        The standard deviation of the computation time for a fold on the valid
+        data.
+    test_time_cv_std : float
+        The standard deviation of the computation time for a fold on the test
+        data.
+    max_ram : float
+        The maximum amount of RAM consume during training.
```
jorisvandenbossche:
```suggestion
        The maximum amount of RAM consumed during training.
```

ramp-database/rampdb/model/submission.py
```diff
@@ -56,8 +56,82 @@
...
 class Submission(Model):
-    """An abstract (untrained) submission."""
-
+    """Submission table.
+
+    Parameters
+    ----------
+    name : str
+        The submission name.
+    event_name : str
```
jorisvandenbossche: the init uses `event_team` ? And maybe it's not a string but an actual EventTeam object?

ramp-database/rampdb/model/event.py
glemaitre: Yes you are right

ramp-database/rampdb/model/event.py
glemaitre: Yes that was it.

ramp-database/rampdb/model/fold.py
glemaitre: Yes

ramp-database/rampdb/model/submission.py
glemaitre: On the top of the file in the enum. It is quite long to state is there I find. But the info is quite missing as well. Do you have a proposal?

ramp-database/rampdb/model/fold.py
glemaitre: Actually they are attached to a submissiononcvfold which are attached to a submission

ramp-database/rampdb/model/submission.py
jorisvandenbossche: just stated that they are stored in that enum

jorisvandenbossche:
Addressed all comments, should be ready

ramp-database/rampdb/model/event.py
```diff
@@ -25,6 +25,88 @@
...
+    private_leaderboard_html : str
+        The private leaderboard in HTML.
+    failed_leaderboard_html : str
+        The leaderboard with the failed submissions.
+    new_leaderboard_html : str
+        The leaderboard with the new submitted submissions.
+    public_competition_leaderboard_html : str
+        The public leaderboard of the competition in HTML.
+    private_competition_leaderboard_html : str
+        The private leaderboard of the competition in HTML.path_ramp_kits : str
```
glemaitre:
```suggestion
        The private leaderboard of the competition in HTML.
```

ramp-database/rampdb/model/event.py
```diff
@@ -25,6 +25,88 @@
...
+        The path where the kits are located.
```
glemaitre:
```suggestion
```

ramp-database/rampdb/model/event.py
```diff
@@ -89,82 +171,129 @@ def __repr__(self):
...
         return len(self.event_teams)
 
 
 class EventScoreType(Model):
+    """EventScoreType table.
+
+    This is a many-to-one relationship between Event and ScoreType. Stores the
+    ScoresTypes for each event.
+    For each Event / ScoreType combo, also a new record in ScoreType is
+    created, which is not that useful (TODO consider removeing ScoreType table)
```
glemaitre:
```suggestion
    created, which is not that useful (TODO consider removing ScoreType table)
```

ramp-database/rampdb/model/fold.py
```diff
@@ -17,9 +17,28 @@
 
 
 class CVFold(Model):
-    """Storing train and test folds, more precisely: train and test indices.
+    """CVFold table.
 
-    Created when the ramp event is set up.
+    Storing train and test folds, more precisely: train and test indices for
+    a single fold. Multiple records of this table are then linked to a
+    submission (through `SubmissionOnCVFold`).
```
glemaitre:
```suggestion
    submission (through :class:`rampdb.model.SubmissionOnCVFold`).
```

ramp-database/rampdb/model/user.py
```diff
@@ -122,7 +202,73 @@ def __repr__(self):
 )
 
 
+# TODO: the ip should not be read from the environment variable but from the
+# config file instead.
+# TODO: Not tested for the moment.
 class UserInteraction(Model):
+    """UserInteraction table.
```
glemaitre:
```suggestion
    """UserInteraction table.
    
    This class is used to record the interaction of a user with the frontend.
```

ramp-database/rampdb/model/user.py
```diff
@@ -122,7 +202,73 @@ def __repr__(self):
 )
 
 
+# TODO: the ip should not be read from the environment variable but from the
+# config file instead.
+# TODO: Not tested for the moment.
```
glemaitre:
```suggestion
```

PR #136, opened 2019-01-31
 MAINT: Consistency of the setup and version across bundle 

glemaitre:


ramp-engine/LICENSE
```diff
@@ -1,29 +0,0 @@
-BSD 3-Clause License
```
jorisvandenbossche: I think we should keep the license in each of the repos?

(in case we package them all separately on pypi (not sure if we plan that), it is nice to have them in the package)

glemaitre: Oh ok, yes we can do that.
We probably need to translate the md to an rst then as well

PR #159, opened 2019-03-06
ENH: provide quickstart commands to create config files

jorisvandenbossche:


jorisvandenbossche:
On the testing (cleaning up the tmp deployment dir, this is actually already handled by the `teardown_function` you added in that file).

This should be good to go I think, or at least to get a look from you.

PR #162, opened 2019-03-06
EHN: remove deployement_dir and use absolute path for RAMP components

glemaitre:


ramp-database/ramp_database/testing.py
```diff
@@ -47,18 +48,35 @@ def create_test_db(database_config, ramp_config):
...
     database_config = database_config['sqlalchemy']
     # we can automatically setup the database from the config file used for the
     # tests.
     ramp_config = generate_ramp_config(ramp_config)
-    shutil.rmtree(ramp_config['deployment_dir'], ignore_errors=True)
+
+    # FIXME: we are recreating the deployment directory but it should be
+    # replaced by an temporary creation of folder.
+    match = SequenceMatcher(
+        None, ramp_config['ramp_kit_dir'], ramp_config['ramp_submissions_dir']
```
jorisvandenbossche: you can use `os.path.commonpath` here (eg between ramp_kit_dir and ramp_data_dir), that would already be much simpler I think. 
But we could also just hard code it, as for the tests we know what the it is (or which part of the path it is, eg do a os.path.split twice on ramp_kit_dir)

ramp-database/ramp_database/tests/data/ramp_config_boston_housing.yml
```diff
@@ -3,12 +3,11 @@ ramp:
...
-    data_dir: ramp-data/boston_housing
-    kit_dir: ramp-kits/boston_housing
-    submissions_dir: submissions
-    predictions_dir: preds
-    logs_dir: log
+    data_dir: /tmp/databoard_test/ramp-data/boston_housing
+    kit_dir: /tmp/databoard_test/ramp-kits/boston_housing
+    submissions_dir: /tmp/databoard_test/submissions
+    predictions_dir: /tmp/databoard_test/preds
+    logs_dir: /tmp/databoard_test/log
```
jorisvandenbossche: question: if you have multiple events, are they all stored in the same directory? (it's not stored in a subdir of this logs_dir path?)

ramp-database/ramp_database/testing.py
glemaitre: oh. i did not search properly. Thanks

ramp-database/ramp_database/tests/data/ramp_config_boston_housing.yml
glemaitre: This is used by the dispatcher. So if you have a config file by event, you can change prediction and log directories

Issue #163, opened 2019-03-06
Validation emails are sent to admin instead of the user


rth:
When a new user signs in and is approved, currently the admin receives the confirmation email instead of the user. I think, the issue is with,

ramp-frontend/ramp_frontend/views/admin.py
```py
        users_to_be_approved = request.form.getlist('approve_users')
        event_teams_to_be_approved = request.form.getlist(
            'approve_event_teams'
        )
        message = "Approved users:\n"
        for asked_user in users_to_be_approved:
            approve_user(db.session, asked_user)

            subject = 'Your RAMP account have been approved'
            body = ('{}, you account have been approved. You can now sign-up '
                    'for any opened RAMP event.'
                    .format(flask_login.current_user.name))
            send_mail(
                to=flask_login.current_user.email, subject=subject, body=body
            )

            message += "{}\n".format(asked_user)
        message += "Approved event_team:\n"
```
where `flask_login.current_user.email` is the admin email, not the user email.

PR #164, opened 2019-03-06
Fix Validation emails are sent to admin instead of the user 

rth:
Closes https://github.com/paris-saclay-cds/ramp-board/issues/163

Tested manually.

There are a few other places where similar change might need to be applied in  `ramp-frontend/ramp_frontend/views/admin.py`


glemaitre:
Thanks Roman!!!

glemaitre:
This one of the part which is not tested fully.

Issue #165, opened 2019-03-06
Re-running failed submission with ramp-launch worker


rth:
I am trying to manually re-run a submission failed due to #154  using,
```
ramp-launch worker --submission <submission_name>
```

and I run into two other issues,
 - [x] the submission name doesn't seem to be validated against the list of exiting submissions. For instance,
  ```
  ramp-launch worker --submission anything --verbose
  ```
  will create a  `submissions/anything` folder then unsurprisingly fail due to,
  ```
  FileNotFoundError: [Errno 2] No such file or directory: '/home/rth/ramp_deployment2/submissions/anything/feature_extractor.py'
  ```
  Not too critical but it might be better to fails with some explicit error message on the CLI.

 - [ ] When using a correct submission name, for a submission that previously failed, I get,
  ```
  ramp-launch worker --submission submission_000000006/ --verbose
2019:03:06 20:12:05 - INFO - RAMP-WORKER - CondaEnvWorker(submission_000000006/): status="setup"
2019:03:06 20:12:05 - INFO - RAMP-WORKER - CondaEnvWorker(submission_000000006/): status="running"


Traceback (most recent call last):
  File "/home/rth/miniconda3/envs/ramp-board/bin/ramp-launch", line 11, in <module>
    load_entry_point('ramp-engine', 'console_scripts', 'ramp-launch')()
  File "/home/rth/src/ramp-board/ramp-engine/ramp_engine/cli.py", line 87, in start
    main()
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
[...]

PR #167, opened 2019-03-06
Add a --force option when running a standalone worker

rth:
Partially addresses https://github.com/paris-saclay-cds/ramp-board/issues/165

Adds a `--force` option to `ramp-launch worker` to delete existing predictions (if any exist). Not sure if this will work fine for the database -- testing now.


glemaitre:
The issue would be that the worker will not push in the database. Workers are agnostic to that. This is actually the job of the dispatcher.

rth:
> The issue would be that the worker will not push in the database. Workers are agnostic to that. This is actually the job of the dispatcher.

Aww I see.

PR #169, opened 2019-03-06
Handle error during setup/launch of worker in Dispatcher

jorisvandenbossche:


glemaitre:
Do we need some test here to check that we will raise the error and stop the worker properly.

jorisvandenbossche:
Probably yes, but for now this only affects the aws one (and I am also not sure for aws how to reliably trigger an error, maybe by giving an non-existing AMI id). 
Is there a way for the conda worker to have a failed set-up ?

glemaitre:
Requesting an environment that does not exist
 

                                                                                                                                      
 

                            
     

lucyleeow:
Address instance not launching properly - so hopefully can close #243

lucyleeow:
ping @maikia @agramfort

PR #173, opened 2019-03-07
Delete predict folder if it exists

rth:
Continues https://github.com/paris-saclay-cds/ramp-board/pull/167

Actually even if run tag a submission to be re-run as suggested in https://github.com/paris-saclay-cds/ramp-board/issues/165#issuecomment-470279119, I still get the error about an existing predict folder in https://github.com/paris-saclay-cds/ramp-board/issues/165#issue-418005998 which crashes the dispatcher.

This removes the folder with result predictions if they exist to avoid it.

cc @glemaitre


glemaitre:
I think it looks fine. This should do the trick. Let's add a regression test afterwards.

Issue #178, opened 2019-03-07
psycopg2.DatabaseError: error with status PGRES_TUPLES_OK and no message from the libpq


rth:
An issue on the frontend that keeps happening now that my RAMP instance sees some traffic. Once it happens once, it applies to all pages that need a database connection.  

Restarting postgres doesn't help. The only solution is to restart the frontend. It starts happening every couple of hours in my case.

```
"GET /problems HTTP/1.1" 500 -
Traceback (most recent call last):
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 2309, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 2295, in wsgi_app
    response = self.handle_exception(e)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 1741, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/rth/src/ramp-board/ramp-frontend/ramp_frontend/views/ramp.py", line 82, in problems
    db.session, interaction='looking at problems', user=user
  File "/home/rth/src/ramp-board/ramp-database/ramp_database/tools/user.py", line 153, in add_user_interaction
    session.commit()
[...]

Issue #151, opened 2019-02-24
psycopg2.OperationalError SSL SYSCALL error: EOF detected


rth:
An issue I recently run into. After starting the frontend with,
```
ramp-frontend launch --host 0.0.0.0
```
the home page works, but I get the following traceback when trying to access `/problems/` URL,
```
176.165.38.156 - - [24/Feb/2019 21:53:17] "GET /problems HTTP/1.1" 500 -
Traceback (most recent call last):
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 2309, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 2295, in wsgi_app
    response = self.handle_exception(e)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 1741, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/rth/src/ramp-board/ramp-frontend/ramp_frontend/views/ramp.py", line 82, in problems
    db.session, interaction='looking at problems', user=user
[...]

PR #180, opened 2019-03-08
 EHN: add gunicorn helper for production deployment 

rth:
This adds a preliminary ginicorn setup as discussed with @glemaitre yesterday.

A number of issues #178 #151 due to the use of `processes>1` with the flask built-in server should be avoided with gunicorn. Also subjectively, I feel like the frontend is much more responsive with this setup.

The server can be started with,
```
cd ramp-frontend
gunicorn -w 8 --bind 0.0.0.0:8080 --access-logfile ./frontend.log 'wsgi:make_app("<config_file_path.yml>")'
```

In terms of API, I'm not convinced it's worth wrapping gunicorn in some  CLI app, and I would very much prefer using the gunicorn CLI directly. The only thing that needs to be passed to the app is the `config.yml` which I agree is done not very cleanly here.

Another limitation of gunicorn is that it doesn't work on Windows.


glemaitre:
> In terms of API, I'm not convinced it's worth wrapping gunicorn in some CLI app, and I would very much prefer using the gunicorn CLI directly. The only thing that needs to be passed to the app is the config.yml which I agree is done not very cleanly here.

We might want just to document it. I agree that there are several choices of a web server that a user can do that might don't want to implement a specific one.

> Another limitation of gunicorn is that it doesn't work on Windows.

For the moment I think that it could be good enough.

Issue #179, opened 2019-03-08
Submission error messages not displayed in  training_error


rth:
Currently, users can't seem to see the error in their submission. For instance,

for the failed run http://ramp.r0h.eu:8080/b14a4b62ab7e3eb3322f098d80ab5b03ebddb6bb/error.txt, the ` training_error` is,
```py
    /home/rth/miniconda3/envs/ramp-solar-wind/lib/python3.7/site-packages/pyarrow/pandas_compat.py:708: FutureWarning: .labels was deprecated in version 0.24.0. Use .codes instead.
      labels = getattr(columns, 'labels', None) or [
    /home/rth/miniconda3/envs/ramp-solar-wind/lib/python3.7/site-packages/pyarrow/pandas_compat.py:735: FutureWarning: the 'labels' keyword is deprecated, use 'codes' instead
      return pd.MultiIndex(levels=new_levels, labels=labels, names=columns.names)
    /home/rth/miniconda3/envs/ramp-solar-wind/lib/python3.7/site-packages/pyarrow/pandas_compat.py:752: FutureWarning: .labels was deprecated in version 0.24.0. Use .codes instead.
      labels, = index.labels
```
while the actual error in `log/submission_000000023/log` is
```py
Traceback (most recent call last):

  File "/home/rth/src/ramp-workflow/rampwf/utils/submission.py", line 111, in train_test_submission
    trained_workflow, X_train)

  File "/home/rth/ramp_deployment2/ramp-kits/solar_wind/problem.py", line 54, in test_submission
    y_proba = self.classifier_workflow.test_submission(clf, X_test_array)

  File "/home/rth/src/ramp-workflow/rampwf/workflows/classifier.py", line 19, in test_submission
    y_proba = clf.predict_proba(X_array)

AttributeError: 'Classifier' object has no attribute 'predict_proba'
```

I am aware there is some traceback obfuscation code but this makes it very difficult for users to understand what went wrong.

https://github.com/paris-saclay-cds/ramp-board/pull/169 is another topic, right?

jorisvandenbossche:
OK, the reason is: we store both the stderr as the output captured by the subprocess in the log file:

https://github.com/paris-saclay-cds/ramp-board/blob/dc3335ded32b410f8680c1ea1bd97923c999d6bf/ramp-engine/ramp_engine/local.py#L143-L151

but then from that method only return stderr, which is what gets stored in the database. So we need to concatenate both in memory as well to return from that method.

[...]

PR #182, opened 2019-03-08
BUG: dispatcher - test that error message is stored correctly in database

jorisvandenbossche:


rth:
Unless you are saying that it's actually not stored in the database, and so it's another issue than https://github.com/paris-saclay-cds/ramp-board/issues/179 ?

jorisvandenbossche:
Yeah, this is not fixing anything (just adding a test I wrote to notice it is already working as expected). See https://github.com/paris-saclay-cds/ramp-board/issues/179 for the explanation.

jorisvandenbossche:
Funny, this is actually failing on travis :-) 
So travis has the same behaviour as we see on Amazon or on your server, but different from our local laptops.

In any case, that means we first have to fix it properly before merging this.

glemaitre:
One more fix!!!

rth:
Thanks for addressing this! Can confirm it resolves https://github.com/paris-saclay-cds/ramp-board/issues/179

PR #186, opened 2019-03-08
FIX: properly test and implement default 

glemaitre:


ramp-utils/ramp_utils/ramp.py
```diff
@@ -29,25 +35,45 @@ def _create_default_path(config, key, path_config):
...
     """
-    if isinstance(config, six.string_types):
-        config = read_config(config, filter_section='ramp')
+    if isinstance(event_config, six.string_types):
+        if (database_config is None or
+                not isinstance(database_config, six.string_types)):
+            raise ValueError(
+                'When "event_config" corresponds to the filename of the '
+                'configuration, you need to provide the filename of the '
+                'database as well, by assigning "database_config".'
```
jorisvandenbossche:
```suggestion
                'database config as well, by assigning "database_config".'
```

ramp-utils/ramp_utils/tests/data/ramp_config_missing.yml
```diff
@@ -0,0 +1,11 @@
+ramp:
+    problem_name: iris
```
jorisvandenbossche: where is this file used?

jorisvandenbossche:
Shouldn't it be possible to now remove the "commonpath hack" ?

glemaitre:
> Shouldn't it be possible to now remove the "commonpath hack" ?

The hack was to go around the removal of `deployment path`. We still do not have this one.
With the default values, it will create the deployment in a weird place. I think that it would be nicer to create temporary deployment path and use this instead of the trick.

Issue #154, opened 2019-03-05
FileNotFoundError when executing submissions


rth:
I am working on the solar wind problem, and to test I submitted the starting kit the few times.

The first one was run fine. The second (and the following ones) errored with the following traceback, 

```
Traceback (most recent call last):
  File "/home/rth/miniconda3/envs/ramp-solar-wind/lib/python3.7/site-packages/rampwf/utils/submission.py", line 96, in train_test_submission
    module_path, X_train, y_train, train_is=train_is)
  File "/home/rth/ramp-deployment/ramp-kits/solar_wind/problem.py", line 43, in train_submission
    module_path, X_df, y_array, train_is)
  File "/home/rth/miniconda3/envs/ramp-solar-wind/lib/python3.7/site-packages/rampwf/workflows/feature_extractor.py", line 11, in train_submission
    feature_extractor = import_file(module_path, self.element_names[0])
  File "/home/rth/miniconda3/envs/ramp-solar-wind/lib/python3.7/site-packages/rampwf/utils/importing.py", line 32, in import_file
    submitted_module = imp.load_source(submitted_path, submitted_file)
  File "/home/rth/miniconda3/envs/ramp-solar-wind/lib/python3.7/imp.py", line 171, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 696, in _load
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 724, in exec_module
  File "<frozen importlib._bootstrap_external>", line 859, in get_code
  File "/home/rth/miniconda3/envs/ramp-solar-wind/lib/python3.7/imp.py", line 157, in get_data
    return super().get_data(path)
  File "<frozen importlib._bootstrap_external>", line 916, in get_data

FileNotFoundError: [Errno 2] No such file or directory: '/home/rth/ramp-deployment/submissions/submission_000000004/feature_extractor.py'
```

which is strange because of that `feature_extractor.py` at that path exists and looks fine. Restarting the dispatcher at the time didn't help.

Today I have restarted both the frontend and the dispatcher again and the job appears to be running, the dispatcher shows,
```
2019:03:05 07:10:27 - INFO - RAMP-DISPATCHER - Worker CondaEnvWorker(submission_000000006): status="running" is still running
```
but in the log I already have the same error for the newly created job..

Have anyone encountered this before? I can provide access to the server if it helps investigating..

[...]

PR #187, opened 2019-03-09
WIP FIX Dispatcher synchronization

rth:
Aims to address https://github.com/paris-saclay-cds/ramp-board/issues/154 

Fix by @glemaitre 

Not sure how to unit-test this. 

With this fix applied once a submission is sent there is a delay of 10-15 s for the server to respond. I'm not sure if it is related.

@glemaitre feel free to take over this PR..

### DIFF
--- a/ramp-frontend/ramp_frontend/views/ramp.py
+++ b/ramp-frontend/ramp_frontend/views/ramp.py
@@ -428,6 +428,8 @@ def sandbox(event_name):
                         src=os.path.join(sandbox_submission.path, filename),
                         dst=os.path.join(new_submission.path, filename)
                     )
+                # we delayed the state of the submission until that we copy the file
+                new_submission.set_state('new')
             except DuplicateSubmissionError:
                 return redirect_to_sandbox(
                     event,
--- a/ramp-frontend/ramp_frontend/views/ramp.py
+++ b/ramp-frontend/ramp_frontend/views/ramp.py
@@ -428,8 +428,11 @@ def sandbox(event_name):
                         src=os.path.join(sandbox_submission.path, filename),
                         dst=os.path.join(new_submission.path, filename)
                     )
-                # we delayed the state of the submission until that we copy the file
+                # we delayed the state of the submission until that we copy
+                # the file
                 new_submission.set_state('new')
+                db.session.commit()
+
             except DuplicateSubmissionError:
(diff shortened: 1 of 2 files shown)


glemaitre:
Yep I think it was a quick hotfix but we should have something better.
I did not have time to commit it yet but basically we should copy the file within the function `add_submission`. Like this the copy will happen before the commit to the database and it will be safe.

Setting the state is a bit brittle. If you are not aware that you set the state of the submission, it could be tricky to know about it.

rth:
> I did not have time to commit it yet but basically we should copy the file within the function add_submission. Like this the copy will happen before the commit to the database and it will be safe

Can we not let the dispatcher create that file instead? That would remove the need for syncronization between the DB and the file system. Related to the discussion in https://github.com/paris-saclay-cds/ramp-board/issues/188

glemaitre:
Yep but we need to store the file into the database. This is a better long term solution I think.

ramp-frontend/ramp_frontend/views/ramp.py
```diff
@@ -428,6 +428,8 @@ def sandbox(event_name):
                         src=os.path.join(sandbox_submission.path, filename),
                         dst=os.path.join(new_submission.path, filename)
                     )
+                # we delayed the state of the submission until that we copy the file
+                new_submission.set_state('new')
```
glemaitre: This function does not commit in the database. If you could either call `db.session.commit()` or use `ramp_database.tools.set_submission(db.session, new_submission.id, state)

glemaitre: Could you do that on your server as well to check if it solves the issue?

rth:
Addressed your comment. 

Though there is an issue with this approach still. Turns out that the submissions that were marked with "check" permanently, were due to the fact that gunicorn worker timed-out (after 30 sec), particularly when the server had some load. Meaning that sending a submission can take more than 30 s which is a lot. 

Now I increased the gunicorn timeout to 1 hour, and this works, but understanding why it takes so long to make a submission with the frontend would be useful.

rth:
FWIW, it's a regular (physical) hard drive, there is no reason why interactions with the file system or writing into the DB would teake that long.

Maybe disabling wrting to the DB for user_interactions would help somewhat.

Issue #190, opened 2019-03-11
Accessing SubmissionOnCVFold by backreference is causing memory leak


rth:
The memory usage by the frontend with some trafic appears to be quite large,

![memory_usage](https://user-images.githubusercontent.com/630936/54160229-49a27700-444f-11e9-8612-5050e0296f90.png)
 
This is using gunicorn with 8 workers, and a large worker timeout (of 1 hours) due to #187 . Reducing the timeout value to a more reasonable amount will likely fix this, but still the amount of memory used is quite surprising. 

After restarting the frontend I see the memory usage go down from 23GB to 7 GB, meaning that 16GB RAM were used by the frontend. Even allowing for multiple processes this value is huge (given that the frontend does not run submissions), and might be worth investigating.

glemaitre:
We should need to profile that to find the cause.
Let's remove the user interaction, they are pretty useless in practice.

jorisvandenbossche:
>  Let's remove the user interaction, they are pretty useless in practice.

Yes, or at least have an option to disable/enable them

kegl:
The apache server at ramp.studio also leaks so this may be an old issue. Pls keep user interactions optionally, with default being switched off.

glemaitre:
The memory leak could be leak to the debug mode of SQLAlchemy:
https://github.com/pallets/flask-sqlalchemy/issues/181

@rth I see that on your server the option is still on. Could you try to turn this off and check again.

rth:
Thanks for investigating! Honestly, I am fairly skeptical about SqlAlchemy leaking memory with default settings. If that was indeed the case that issue would see much more activity. Also we are talking about GB of memory, that is a lot of debug information for a website with 20 visitors.

I'll try disabling the debug parameter at some point, but another avenue to explore is,

> If your application is calling create_engine() many times, like per test, or per request, or something like that, this will cause ORM related caches to grow to their max size.

Are we sure we are not re-creating a new engine for each request? Reusing the DB connection might help https://stackoverflow.com/a/17998485/1791279

glemaitre:
We are only creating a single engine. Then, I don't know how it is managed with multiple processes.

rth:
Hmm, yes you are right, it only happens when the app is started.

I imagine each process has its own engine which should be fine.

glemaitre:
But this is true that we don't implement any teardown anywhere.
I am not really sure where it should go and it would work.

kegl:
[...]

PR #192, opened 2019-03-13
MAINT: remove user interaction

glemaitre:
@rth  Could you check if you still have some memory leak without tracking user interaction?

### DIFF
--- a/ramp-frontend/ramp_frontend/tests/test_admin.py
+++ b/ramp-frontend/ramp_frontend/tests/test_admin.py
@@ -239,13 +239,13 @@ def test_update_event(client_session):
         assert event.min_duration_between_submissions == 0
 
 
-def test_user_interactions(client_session):
-    client, _ = client_session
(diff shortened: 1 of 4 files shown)


rth:
I think user interactions just add transactions, so by disabling them, we will just make the leak less visible not really resolving it. cf related comment in https://github.com/paris-saclay-cds/ramp-board/issues/190#issuecomment-472862249

glemaitre:
Yes, I still don't know where to specify where the option should be available.

rth:
Maybe some environment variable as a start,
```py
def add_user_interaction(....):

    if not os.environ.get('RAMP_USER_INTERACTIONS', False):
        return None
    
```
?

rth:
One reason I think interactions might be good to have is that, doing a load test (with multiple requests) on say just `/problems` should stress the database somewhat and see if we can reproduce the memory issues in #190 without doing any complicated actions.

PR #196, opened 2019-03-14
MAINT Remove ramp_frontend/security

rth:
This removes `ramp_frontend/security.py` that is never used and relies on an import `ramp_frontend.app` that is not valid.

Unless you are planning to keep this functionality?


rth:
Also the required `isitdangerous` is not a direct (or indirect) dependency...

glemaitre:
I think that it was used in the form but I don't remember the reason.
But yes let's do that. The tests are passing.

PR #197, opened 2019-03-15
[WIP] BUG: fix deployment using relative path in configuration

glemaitre:
I introduced a bug when allowing for relative path in the configuration.

TODO:

- [x] It is difficult to test the relative import. I think an integration test testing the full process of deploying a ramp kits (iris) with relative import.
- [ ] Solve the issue with CLI worker


glemaitre:
@jorisvandenbossche If you could have a look to the CLI of the worker.
We should change the call to this function. I temporary pass the event_config but it should not be the case. I think that for this worker, we should actually pass all the information without any config.

glemaitre:
Merging what opening an issue for the remaining one

PR #199, opened 2019-03-17
BUG: avoid to update the whole leaderboard when adding a submission

### LINKED ISSUE #190: Accessing SubmissionOnCVFold by backreference is causing memory leak

glemaitre:
closes #190

### DIFF
--- a/ramp-database/ramp_database/tools/leaderboard.py
+++ b/ramp-database/ramp_database/tools/leaderboard.py
@@ -331,7 +331,7 @@ def get_leaderboard(session, leaderboard_type, event_name, user_name=None,
     return df_html
 
 
-def update_leaderboards(session, event_name):
+def update_leaderboards(session, event_name, pending_only=False):
     """Update the leaderboards for a given event.
 
     Parameters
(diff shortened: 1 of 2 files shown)


ramp-database/ramp_database/tools/leaderboard.py
```diff
@@ -377,21 +383,26 @@ def update_user_leaderboards(session, event_name, user_name):
         The event name.
     user_name : str
         The user name. If None, scores from all users will be queried.
+    pending_only : bool, default is False
+        Whether or not to update the whole leaderboards or only the pending
+        submissions. You can turn this option to True when adding a new
+        submission in the database.
```
jorisvandenbossche: It's not fully clear (from reading this PR) what this exactly means. What are "pending" submissions? (those that not have been trained yet?)

glemaitre: The new one only. We can change the name

PR #212, opened 2019-03-18
EHN: add an option to disable/enable the user interaction

glemaitre:
Add an option to disable the user interaction when launching the front-end.


ramp-utils/ramp_utils/template/database_config.yml
```diff
@@ -16,6 +16,7 @@ flask:
     # mail_use_tls: false
     # mail_use_ssl: false
     # mail_debug: false
+    track_user_interaction: True
```
jorisvandenbossche: default to False?

glemaitre:
I see it as a debug mode. So I think it is safer with false. 

PR #216, opened 2019-03-18
DOC: add info to derive your own worker

glemaitre:
Some documentation to show how to build your own worker

### DIFF
--- a/doc/workers.rst
+++ b/doc/workers.rst
@@ -1,3 +1,5 @@
+.. _all_workers:
+
 Notes regarding the RAMP workers
 ================================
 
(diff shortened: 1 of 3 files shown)


doc/workers.rst
```diff
@@ -25,9 +27,33 @@ Available workers:
...
+<https://github.com/ramp-kits/iris>`_, the starting kit requires numpy, pandas,
+and scikit-learn. Thus, the environment used to run the initial submission
+should have these libraries installed.
+
+Subsequently, you can create such environment and use it in the event
+configuration::
 
-Running a submissions on Amazon Web Services (AWS)
---------------------------------------------------
+      conda create --name ramp-iris numpy pandas scikit-learn
```
jorisvandenbossche: I think ramp-workflow is missing here? 
So it will not be a one liner to have the environment

glemaitre: true

PR #228, opened 2019-03-21
MAINT: move some utils from ramp-utils to ramp-database

glemaitre:


jorisvandenbossche:
Is the `import_module_from_source` potentially something to use in ramp-workflow as well? (which would be a reason to keep it in ramp-utils)

glemaitre:
Yes it is. Another solution is to duplicate. 
We can let it now for the moment and see if we can remove the ramp utils

jorisvandenbossche:
ramp-utils has the core cli (which we should add as dependency for ramp-workflow at some point), so that's a reason to keep it?

glemaitre:
yep until ramp-utils can be installed in standalone for ramp-workflow.
I think that we can keep as it is now.

PR #230, opened 2019-03-27
DOC: add path information when deploying RAMP

glemaitre:
It could be easy to get confused with the different path.
I made more explicit explanations regarding the path where the different commands are run.

### DIFF
--- a/doc/create_ramp_event.rst
+++ b/doc/create_ramp_event.rst
@@ -9,21 +9,22 @@ Deploy a specific RAMP event
 Now we will want to deploy a specific problem and event. We will take an
 example by deploying an ``iris`` event.
 
-First, you need to get the starting kit and the data::
+First, you need to get the starting kit and the data within the
+``ramp_deployment`` directory::
 
-    mkdir ramp-kits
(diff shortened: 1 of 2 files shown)


doc/create_ramp_event.rst
```diff
@@ -9,21 +9,22 @@ Deploy a specific RAMP event
...
-    git clone https://github.com/ramp-data/iris ramp-data/iris
-    cd ramp-data/iris
-    python prepare_data.py
-    cd ../..
+    ~/ramp_deployment $ mkdir ramp-kits
+    ~/ramp_deployment $ mkdir ramp-data
+    ~/ramp_deployment $ git clone https://github.com/ramp-kits/iris ramp-kits/iris
+    ~/ramp_deployment $ git clone https://github.com/ramp-data/iris ramp-data/iris
+    ~/ramp_deployment $ cd ramp-data/iris
+    ~/ramp_deployment/ramp-kits/iris $ python prepare_data.py
```
jorisvandenbossche: this should be ramp-data instead of ramp-kits ?

PR #236, opened 2019-03-29
EHN: force user to confirm their email address

glemaitre:
This PR force a user to confirm his email address:

* force user to have a real email address;
* remove from the list to approve email which has not been confirmed.


ramp-frontend/ramp_frontend/views/auth.py
```diff
@@ -145,24 +145,34 @@ def sign_up():
...
+        body += '\n\nSee you on the RAMP website!'
+        send_mail(user.email, subject, body)
+        logger.info(
+            '{} has signed-up to RAMP'.format(user.name)
+        )
+        flash(
+            "We sent a confirmation email. Go read your email and click on "
+            "the confirmation link"
+        )
+        print(body)
```
glemaitre: to be removed

PR #251, opened 2019-04-03
EHN: add more command lines to interact with database

### LINKED ISSUE #249: Add more command line option for dealing with database
All the current option are not wrapped within the CLI api which can be a problem sometimes.

glemaitre:
closes #249 

An example of what I meant in #249.
I let this opened to be taken back.


rth:
This needs a test before we can merge it though.

PR #254, opened 2019-04-04
BUG directly log into file for local worker

### LINKED ISSUE #287: Some workers hang when stderr is too long
When some jobs trow a lot of warnings, this flood the stderr pipe and the job cannot return at the end. To solve this, the dispatcher should redirect stdout and stderr directly in files on the disk.

glemaitre:
closes #287 

subprocess.PIPE might block the dispatcher. Instead we are logging directly in a file.


glemaitre:
@agramfort @lucyleeow @maikia We absolutely need this one as well.

agramfort:
tell me when good to go @glemaitre

Issue #259, opened 2019-10-25
test_check_bagged_scores fails with pandas 0.25+


rth:
Using the latest version of dependencies used by default in [environment.yml](https://github.com/paris-saclay-cds/ramp-board/blob/master/environment.yml) (at present pandas 0.25.2) `test_check_bagged_scores` fails with,
```
______________________________________________________ test_check_bagged_scores ________________________________
_______________________

session_scope_module = <sqlalchemy.orm.session.Session object at 0x7fe3898b91d0>

    def test_check_bagged_scores(session_scope_module):
        # check both set_bagged_scores and get_bagged_scores
        submission_id = 1
        path_results = os.path.join(HERE, 'data', 'iris_predictions')
        set_bagged_scores(session_scope_module, submission_id, path_results)
        scores = get_bagged_scores(session_scope_module, submission_id)
        multi_index = pd.MultiIndex(levels=[['test', 'valid'], [0, 1]],
                                    labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
                                    names=['step', 'n_bag'])
        expected_df = pd.DataFrame(
            {'acc': [0.70833333333, 0.70833333333, 0.65, 0.6486486486486],
             'error': [0.29166666667, 0.29166666667, 0.35, 0.35135135135],
             'nll': [0.80029268745, 0.66183018275, 0.52166532641, 0.58510855181],
             'f1_70': [0.66666666667, 0.33333333333, 0.33333333333,
                       0.33333333333]},
            index=multi_index
        )
        expected_df.columns = expected_df.columns.rename('scores')
>       assert_frame_equal(scores, expected_df, check_less_precise=True)
ramp_database/tools/tests/test_submission.py:401:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
_ _ _ _ _ _ _ _ _ _ _ _
pandas/_libs/testing.pyx:65: in pandas._libs.testing.assert_almost_equal
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[...]

PR #260, opened 2019-10-25
Pin pandas version to 0.24.x

rth:
A temporary workaround for https://github.com/paris-saclay-cds/ramp-board/issues/259 by pinning pandas to 0.24.x

Starting with an empty commit to see if I can reproduce the issue in CI.


rth:
Closing as following discussion https://github.com/paris-saclay-cds/ramp-board/issues/259 probably better to fix this properly.

PR #268, opened 2019-11-06
Update install.rst

kegl:
Adding necessary command to run tests.

ramp-database and ramp-utils are fine now, engine and frontend still fail. It would be nice to fix this.

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -83,7 +83,17 @@ Test
 ====
 
 If you are developing for the RAMP package, you will be interested about
-testing your new feature. You can run the test using ``pytest`` from the root
+testing your new feature. First you need to `set up the database <https://paris-saclay-cds.github.io/ramp-docs/stable/setup_server.html#set-up-the-ramp-database>`_ and create the test database::
+
+    createdb --owner=<the owner you specified> databoard_test
+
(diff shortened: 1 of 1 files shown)


kegl:
Actually ramp-engine is fine if I pip install it in development mode.

doc/install.rst
```diff
@@ -83,7 +83,17 @@ Test
 ====
 
 If you are developing for the RAMP package, you will be interested about
-testing your new feature. You can run the test using ``pytest`` from the root
+testing your new feature. First you need to `set up the database <https://paris-saclay-cds.github.io/ramp-docs/stable/setup_server.html#set-up-the-ramp-database>`_ and create the test database::
```
agramfort: this is not the right way to link within sphinx. you need to link to local build.

agramfort: a dev build of the website would point to stable

glemaitre:
I don't think that this test section should be placed in the `install.rst`. After some feedback from @lucyleeow and @maikia, it seems that this section should be in a `maintainer.rst` or something similar. The reason is that we are expecting specific username and password to test on the database and people get confused with the part of the deployment then.

doc/install.rst
```diff
@@ -83,7 +83,17 @@ Test
...
+
+    createdb --owner=<the owner you specified> databoard_test
```
glemaitre: Here you will not have any choice. The config that we are testing thing require `mrramp` as password and user.

Issue #193, opened 2019-03-14
Discrepancy between the train time and evaluation time


rth:
The training times shown on RAMP don't seem to be consistent with the evaluation time.

For instance, for the solar_wind problem, the starting kit says that the train time is 0.54 s (and test is 0.17 s), while to evaluate that submission it takes ~25min on the same server.

The evaluation time is consistent with what I get by running the `ramp_test_submission` locally, and the reported train/test time is faster by 3000x  which doesn't make sense even allowing for multiple CV folds.

rth:
As a side note, @glemaitre  don't you think that having a "Frontend" etc. labels would be easier to manage that renaming each issue :)

glemaitre:
I feel stupid now :) I like only to use 10% of GitHub ;)
I will create those.

glemaitre:
Be aware that the reported time for training is only on a single fold. So you need to multiply by the number of folds. However, x3000 seems kinda of unlikely :)

kegl:
You also need to score which is slow sometimes.

glemaitre:
Uhm we solve this issue. We were reporting the validation time instead of the training time.

PR #117, opened 2019-01-14
MAINT/TST/DOC: backend/frontend refactoring

glemaitre:
TODO:

- [x] Unit test for the leaderboard
- [x] Send email with the frontend

closes #114 
closes #109 
closes #110 
closes #111
closes #107 
closes #100 
closes #96 
closes #92

PR #276, opened 2019-11-07
Changes for the acrobot hackathon

kegl:
* [x] Optional credit tracking, set by the "TRACK_CREDITS" config variable. Credit tracking is needed if we want to compute historical contributivities.  *PR in https://github.com/paris-saclay-cds/ramp-board/pull/449*
* [x] CLI command `ramp-database compute-contributivity --event --ramp-kit-dir --ramp-data-dir`. It computes contributivity and historical contributivity scores (displayed in two leaderboard columns) and the combined score (displayed at the top of the leaderboard pages). It is not run automatically as before, admins need to update it time to time. Running time is quadratic in number of submissions, and can be quite long of the data sets are big (more precisely: if the combination function is slow). *PR in https://github.com/paris-saclay-cds/ramp-board/pull/449*
* [x] CLI command `ramp database make-admin --user` to make existing user site admin. *Added in https://github.com/paris-saclay-cds/ramp-board/pull/308*
* [ ] Accelerating leaderboard computation when predictions are big.
* [x] When multiple submissions are processed in the dispatcher.update_database_results, we only update the leaderboard once (not for each processed submission). *Fixed in https://github.com/paris-saclay-cds/ramp-board/pull/316*
* [x] `ramp_database.tools.submission.get_scores` and `get_time` accelerated by avoiding loading predictions from SubmissionOnCVFold. (related ordering bug fixed in https://github.com/paris-saclay-cds/ramp-board/pull/321) *Solved by https://github.com/paris-saclay-cds/ramp-board/pull/315*
* [x] Fixing the leaderboard train time (it displayed validation time) *Solved https://github.com/paris-saclay-cds/ramp-board/issues/193 (likely in https://github.com/paris-saclay-cds/ramp-board/pull/117))*
* [ ] beautifying the leaderboards (new order in private, rounding integers, renaming some columns)
* [x] Toggling submissions in and out of the competition added back. The frontend button was kept but there was no view function catching it. It didn't crash, just didn't do anything. *Solved in https://github.com/paris-saclay-cds/ramp-board/pull/351*
* [x] In the collaborative phase participants look at each other's code. There was a crash since a query tried to find the submission using the current user's name, not the submitter's name. Hotfix is done by returning True if the event is public, before querying the submission. Not yet satisfactory though, minor bug remains in code. *Fixed in https://github.com/paris-saclay-cds/ramp-board/pull/334*


kegl:
@glemaitre: does the CLI part of the sphynx doc get updated automatically as we add CLI commands or it has to be added manually to the doc?

kegl:
I prefer not to for readability.

kegl:
@agramfort, @glemaitre I prefer to elaborate to avoid misunderstandings although it's a minor issue.

The new version is about 30 lines longer, 20 of which is docstring and comments. We could add helper functions to get the scores and the time and call it from both the private and public leaderboard code. That would be the proper way to modularize, but that would add code (new functions, docstrings, etc). I'm not against but I find this current compromise (between code repetition and length) better.

The previous code somehow assumed that there are only minor differences between public and private leaderboard and tried to construct them in the same function. The timing bug was introduced by an over-generalizing idea to make it "pretend" that they are more similar than they actually are. Even tests didn't get this bug. It was quite important since in our challenge there is limit on total _training_ time, so we needed a hotfix in the weekend. When doing it I had a hard time understanding the code, quite a bit of pandas magic. I prefer to construct the columns explicitly, for readability. In this way we can also modify the the two tables relatively independently.

kegl:
I would like to merge this into master after the hackathon ends, so early December.

Hopefully no more hotfixes, we'll see.

We have one more customization change pending: we added a long privacy statement and terms and conditions on the sign up page (required by our lawyers). We simply modified the sign up template, but it's pure html that could be programmatically inserted from a file into the template. This should be customizable. @agramfort @glemaitre @maikia @lucyleeow what would you suggest to do here? We can have custom html files in https://github.com/paris-saclay-cds/ramp-board/tree/master/ramp-frontend/ramp_frontend/templates but is this a good practice to modify those directly (since they are part of the software, committed in the repo)?

rth:
I think it would help a lot reviewers if this PR was split in smaller chunks. I have not reviewed everything in details, but following parts should be very easy to merge if they were made as separate PRs and looks like a significant improvement,
 - optimization in `ramp-database/ramp_database/tools/frontend.py` and in dispatcher `ramp-engine/ramp_engine/dispatcher.py`
 - fixing typos in `ramp_frontend/views/admin.py` and `ramp_frontend/views/auth.py`
 - fix bug in `send_mail(admin`: I have actually run into this previously, but hasn't opened an issue so far.
 - the addition of make_admin CLI command

and in larger changes (that I have not reviewed in detail),
 - adding the computation of contribution
 - and other CV changes discussed in the description

If anyone interested/available in making smaller PRs for the top 4 points, at least? That would also help reducing the diff here. Maybe @maikia ? :)

rth:
Closing as I believe all major features from this PR were either merged as separate PRs to master or are now in the advanced branch. Full list in the issue description https://github.com/paris-saclay-cds/ramp-board/pull/276#issue-338033276
There might be a few style related points remaining but for those it would be easier to create a new than trying to resolve merge conflicts there IMO.

PR #280, opened 2019-11-12
[MRG] Create database for tests

### LINKED ISSUE #257: test db required before pytest
It seems to me that we need to create the db user "mrramp" and database "databoard_test" before launching pytest. Maybe we should mention this in the doc?

maikia:
closes #257 

Testing database locally requires test database which needs to be started manually before starting the tests. To simplify the test process I added the conftest.py with a fixture which starts postgres database, creates admin user and creates the test database before the tests and destroys it afterwards.

#### Any other comments?
the frontend tests fail on the smtp connection. I am trying to add another fixture to create smtp connection before the tests. If you think it is useful I would appreciate comments here @kegl , @glemaitre


agramfort:
great initiative @maikia ! thx a lot

maikia:
After discussion and help of @glemaitre we decided to:

1) add setup.cfg with username and password used to connect to the database. (this will need to come with small change in the documentation)
add it to .gitignore for use of travis with connection setup for travis: postgresql://postgres:@localhost/postgres

2) not make smtp server fixture but add decorator to smtp server tests (@pytest.mark.skipif the connection does not exist) so that they don't give the local error. After finishing conftest I will deal with this separately

.travis.yml
```diff
@@ -17,8 +17,8 @@ install:
     - source activate testenv
     - make inplace
 script:
-    - psql -U postgres -c "CREATE USER mrramp WITH PASSWORD 'mrramp';ALTER USER mrramp WITH SUPERUSER;"
-    - createdb --owner=mrramp databoard_test
+    - psql -l
```
glemaitre: You can remove these lines

ci_tools/travis/integration_test.sh
```diff
@@ -4,6 +4,9 @@ cd $HOME
 mkdir ramp_deployment
 cd ramp_deployment
 ramp setup init
+psql -U postgres -c "CREATE USER mrramp WITH PASSWORD 'mrramp';ALTER USER mrramp WITH SUPERUSER;"
```
glemaitre: These commands should come before the `ramp setup init`, isn't it?

conftest.py
```diff
@@ -0,0 +1,49 @@
...
+import smtpd
+from sqlalchemy import create_engine, exc
+from threading import Thread
+
+from yaml import safe_load
+
+
+@pytest.fixture(scope='session')
+def database_connection():
+    '''
```
glemaitre: double quotes are more in line with numpydoc

```python
"""Create a Postgres database for the tests and remove it
when the tests are done.
```

conftest.py
```diff
@@ -0,0 +1,49 @@
...
+    Create a Postgres database for the tests,
+    and drop it when the tests are done.
+    '''
+    #os.system('pg_ctl -D postgres -U postgres -l logfile start')
```
glemaitre: You probably want to remove the commented line

conftest.py
```diff
@@ -0,0 +1,49 @@
...
+
+    #engine = create_engine('postgresql://<local_user>:@localhost/<engine_name>', 
+    #                       isolation_level='AUTOCOMMIT')
+    config = safe_load(open("db_engine.yml"))
+    dbowner = config.get('db_owner')
+    dbcluster = config.get('db_cluster_name')
+    engine = create_engine(f'postgresql://{dbowner}:@localhost/{dbcluster}', 
+                           isolation_level='AUTOCOMMIT')
+    #engine = create_engine('postgresql://postgres:@localhost/import argparse',
```
glemaitre: remove this one

conftest.py
```diff
@@ -0,0 +1,49 @@
...
+    #                       isolation_level='AUTOCOMMIT')
+    connection = engine.connect()
+
+    try:
+        connection.execute("""CREATE USER mrramp WITH PASSWORD 'mrramp';
+                              ALTER USER mrramp WITH SUPERUSER""")
+    except exc.ProgrammingError:
+        print('mrramp already exists. Working with existing mrramp')
```
glemaitre: I would raise an error instead. Imagine that somebody create a database with these credentials, we will destroy it.
Maybe it is just safer.

conftest.py
```diff
@@ -0,0 +1,49 @@
...
+
+    try:
+        connection.execute('CREATE DATABASE databoard_test OWNER mrramp')
+    except exc.ProgrammingError:
+        print('database exists. Reusing existing database')
```
glemaitre: The same here.

maikia:
Great! thanks for all your comments. Yes, I need to add the explanation to the docs.
For now the user needs to manually create db_engine.yml in ramp-board directory (the one which I put in git comes with the settings for travis). 
Do you agree with this way or do you think of a better way to add this file?

maikia:
I will make docs in a separate PR, otherwise I think it's ready.
Could you please have another look @glemaitre ?
Thanks!

maikia:
Thanks @glemaitre and @agramfort

PR #286, opened 2019-11-20
[MRG] Ovh logo

maikia:
Added Ovh logo to the front webpage and removed other powered-by logos


kegl:
We need to find a way to customize some of the web pages (landing, sign up). Logos is one thing, but for example we needed to add terms and conditions to the sign up page. I included it in the template for now as plain html, but I will not commit it in the master. Since these may come up later, what is important is to find a way to do this seamlessly. Have placeholders and defaults for those committed in the master, and a config template folder in the deployment where customizable elements can be changed. Then we can copy those into the template directory (ignored by github) when for example the frontend is launched.

agramfort:
@maikia can you see if it's possible to make it a configurable thing?

>

maikia:
> @maikia can you see if it's possible to make it a configurable thing?
> […](#)

@kegl @agramfort 
Sure, I do my best. Just to be clear as I never worked with placeholders and defaults (or html for that matter ;) )

1) you want the frontpage to be easily customizable (logos, pictures, terms and conditions, sections, etc) before the launch?
2) How about after the launch? If like now we need to change logo(s) (or people, or other info) on the running RAMP... Is there a way and if yes is it of interest to have possibility of making some changes like that to already launched RAMP?
3) finally, how much of a priority is it?

agramfort:
I would do this with the config file of ram-board. It could contain the
paths to logos to use for acknowledgements.

medium priority I would say.

kegl:
Ideally we should be able to change it after the lunch. I simply edited the sign up template https://github.com/paris-saclay-cds/ramp-board/blob/master/ramp-frontend/ramp_frontend/templates/sign_up.html adding pure html to it. You can do the same thing temporarily at your end to add logos. The problem is that sign_up.html is in the repo and I don't want to commit Huawei T&C. The same way, I don't think OVH logo should be committed in the repo. 

I would set up a deploy_templates folder in the deployment folder in which some default customizable templates can be placed. These could be copied into ramp-frontend/ramp_frontend/templates at deployment and re-copied any time we change them. These could be then sent to index or sign_up through the flask code. I think we don't even have to restart the frontent since the flask could read these files dynamically.

@maikia if you set this up for the logo section of index.html, I can take care of the T&C on the sign_up page.

maikia:
@kegl @agramfort 
Now the logos are taken directly from static/img/powered_by/ directory. No more hardcoded logos
If that's what you wanted I will add it to the docs. Not sure how to add a test for it though.

agramfort:
nice, simple and very elegant !

+1 for MRG

feel free to MRG @kegl

kegl:
Nice! Maybe add the folder to .gitignore and add documentation.

maikia:
I think it's better to .gitignore the contents of the powered_by directory and not of the directory itself

agramfort:
good to go form my end

kegl:
Great job @maikia, +1 for MRG

maikia:
Great, thanks @agramfort  @kegl

ramp-frontend/ramp_frontend/views/general.py
```diff
@@ -11,7 +12,12 @@
 @mod.route('/')
 def index():
     """Default landing page."""
-    return render_template('index.html')
+    img_ext = ('.png', '.jpg', '.jpeg', '.gif', '.svg')
+    images = [f for f in os.listdir(
+              'ramp-frontend/ramp_frontend/static/img/powered_by/')
```
rth: This path should be with respect to `__file__`, otherwise it does not exist https://github.com/paris-saclay-cds/ramp-board/issues/300

ramp-frontend/ramp_frontend/views/general.py
```diff
@@ -11,7 +12,12 @@
...
+              if f.endswith(img_ext)]
+    print(images)
```
rth: Also maybe remove the print?

PR #293, opened 2019-12-02
[WIP] Check if use has been approved

lucyleeow:
Checks if user has been approved.

Need to include test to check that sign in button no longer appears after a user has signed up for an event but not yet been approved. Not sure how to go about this - is the [`update_event`](https://github.com/paris-saclay-cds/ramp-board/blob/1ba84e23b92186f2a5b998596a8c8a116c24d9ff/ramp-frontend/ramp_frontend/views/admin.py#L164) function generating the html for event pages?
Should the test use [`client_session`](https://github.com/paris-saclay-cds/ramp-board/blob/1ba84e23b92186f2a5b998596a8c8a116c24d9ff/ramp-frontend/ramp_frontend/tests/test_admin.py#L29) to create a user session?

### DIFF
--- a/ramp-database/ramp_database/tools/frontend.py
+++ b/ramp-database/ramp_database/tools/frontend.py
@@ -140,3 +140,28 @@ def is_user_signed_up(session, event_name, user_name):
             (event_team.is_active and event_team.approved)):
         return True
     return False
+
+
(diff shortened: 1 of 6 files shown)


glemaitre:
closing this one since it has been merged already

PR #295, opened 2019-12-03
ENH Add depth parameter to setup_ramp_kit_ramp_data

rth:
Ramp starting kit repos can be somewhat large due to the history of the jupyter notebooks they contain.

This adds an optional `depth` parameter to `setup_ramp_kit_ramp_data` that is going to be passed to `git clone`. In particular, `depth=1` will make a shallow clone (used in particular in tests by default).

Not sure this matters much in practice, but tests are slow so I imagine anything helps..


glemaitre:
Thanks @rth

Issue #189, opened 2019-03-11
Add worker timeout


rth:
Currently, it doesn't look like workers timeout by default. I see submissions that are still running after 18 hours, and that would never finish because the submitted algorithm just doesn't scale.

It would be good to have some timeout mechanism for workers; or if it exists enable it by default to kill running jobs after say 6 hours.

glemaitre:
This a good option to add.

PR #296, opened 2019-12-04
ENH Add conda worker timeout

rth:
This implements timeout for conda workers. This actually wont be solved by running in a Docker container and so needs to be implemented separately.

Closes https://github.com/paris-saclay-cds/ramp-board/issues/189

Timeout can be specified in event config e.g.

```yaml
worker:
    worker_type: conda
    conda_env: air_passengers
    timeout : 3600  # in seconds
```
with the default value of `7200` (2 hours). Maybe that value should be increased a bit. 

This PR could probably use more tests. In particular on the dispatcher side.

cc @tomMoral @glemaitre


rth:
The LGTM warning looks like a false positive...

ramp-engine/ramp_engine/local.py
```diff
@@ -162,6 +183,10 @@ def collect_results(self):
             if os.path.exists(pred_dir):
                 shutil.rmtree(pred_dir)
             shutil.copytree(output_training_dir, pred_dir)
+            if self.status == 'timeout':
+                returncode = 124
```
rth: Same return code as the `timeout` posix command...

ramp-engine/ramp_engine/base.py
```diff
@@ -67,6 +67,10 @@ def status(self):
     def status(self, status):
         self._status = status
 
+    def check_timeout(self):
```
glemaitre: Should we make it an abstract method or could we implement something to be shared between the different worker type (the one in `aws/worker.py`)

ramp-engine/ramp_engine/local.py
```diff
@@ -127,6 +144,7 @@ def launch_submission(self):
             stderr=subprocess.PIPE
         )
         super().launch_submission()
+        self._start_date = datetime.utcnow()
```
glemaitre: for instance, this could be done in the base class I think

ramp-engine/ramp_engine/tests/test_conda_worker.py
```diff
@@ -168,3 +170,32 @@ def test_conda_worker_error_soon_collection(get_conda_worker):
...
+        worker.setup()
+        assert worker.status == 'setup'
+        worker.launch_submission()
+        assert not worker.check_timeout()
+        assert worker.status == 'running'
+        sleep(2)
+        assert worker.check_timeout() is True
+        assert worker.status == 'timeout'
+        exit_status, _ = worker.collect_results()
+        assert exit_status > 0
```
glemaitre: we can make sure that we raise `124`

kegl:
Does this kill the worker or the training process of the particular submission?

glemaitre:
Since inheritance is not that easy between worker here, let's merge this PR and open another issue for the AWS worker.

rth:
Thanks @glemaitre !

Issue #181, opened 2019-03-08
select_submission_by_name: sqlalchemy.orm.exc.NoResultFound: No row was found for one


rth:
Another exception I saw in the frontend logs,

```
Submission starting_k already exists. Please change the name.
You need to wait 819 more seconds until next submission
[2019-03-07 22:53:47 +0000] [17904] [ERROR] Error handling request /64bece7bf958d9621b4e29a475759dfc68f85374/feature_extractor.py
Traceback (most recent call last):
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/gunicorn/workers/sync.py", line 135, in handle
    self.handle_request(listener, req, client, addr)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/gunicorn/workers/sync.py", line 176, in handle_request
    respiter = self.wsgi(environ, resp.start_response)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 2309, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 2295, in wsgi_app
    response = self.handle_exception(e)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 1741, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/rth/miniconda3/envs/ramp-board/lib/python3.7/site-packages/flask/_compat.py", line 35, in reraise
    raise value
[...]

PR #298, opened 2019-12-05
FIX Use query(...).one_or_none in ramp_database._query

rth:
Closes https://github.com/paris-saclay-cds/ramp-board/issues/181

Generally use `query(...).one_or_none()` instead of `query(..).one()` to avoid exceptions when no results exist in `select_submission_by_name` and `select_user_by_name`.  + some tests.

### DIFF
--- a/ramp-database/ramp_database/tools/_query.py
+++ b/ramp-database/ramp_database/tools/_query.py
@@ -93,7 +93,7 @@ def select_submission_by_name(session, event_name, team_name, name):
                    .filter(EventTeam.id == Submission.event_team_id)
                    .filter(Submission.name == name)
                    .order_by(Submission.submission_timestamp)
-                   .one())
+                   .one_or_none())
 
 
 def select_event_by_name(session, event_name):
@@ -158,7 +158,7 @@ def select_user_by_name(session, user_name):
     """
     if user_name is None:
         return session.query(User).all()
-    return session.query(User).filter(User.name == user_name).one()
+    return session.query(User).filter(User.name == user_name).one_or_none()
 
 
 def select_user_by_email(session, email):
--- a/ramp-database/ramp_database/tools/tests/test_query.py
+++ b/ramp-database/ramp_database/tools/tests/test_query.py
@@ -1,0 +1,88 @@
+import pytest
+import shutil
+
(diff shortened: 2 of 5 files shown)


kegl:
I'm not completely sure about this, I'd prefer raising the exception up with a meaningful error message. Calling these functions with non-existing objects indicates some kind of bug.

rth:
Thanks for the review.

> Calling these functions with non-existing objects indicates some kind of bug.

Not necessarily. For a variety of reasons calling `select_submission_by_name` or `select_user_by_name` can return no results. Raising exceptions in that case, means that one has to always wrap these calls into `try: ... except: ` clauses, leading to complicated nested try except code e.g. in, https://github.com/paris-saclay-cds/ramp-board/blob/4bc503b350343f6f34617b28d8e96e06c534cdc5/ramp-database/ramp_database/tools/user.py#L90-L100

Returning `None` when the entry does not exist, means one still has to check for it, and an exception would be raised if one tries to do anything with it (and the traceback should be understandable). However it makes the code much more readable, here,
``` py
		if select_user_by_name(session, name) is not None:
            message += 'username is already in use'
        elif select_team_by_name(session, name) is not None:
            message += 'username is already in use as a team name'
```

rth:
Also  most of the helper functions in `tools._query` already use `one_or_none()` on master,
```
$ rg "\.one()" ramp-database/ramp_database/tools/_query.py                                  [master] 
96:                   .one())
116:    return session.query(Event).filter(Event.name == event_name).one_or_none()
141:                   .one_or_none())
161:    return session.query(User).filter(User.name == user_name).one()
179:    return session.query(User).filter(User.email == email).one_or_none()
199:    return session.query(Team).filter(Team.name == team_name).one_or_none()
221:                   .one_or_none())
287:                   .one_or_none())
309:                   .one_or_none())
332:                   .one_or_none())
367:    return q.one_or_none()
392:                   .one_or_none())
413:                   .one_or_none())
439:                   .one_or_none())
462:                   .one_or_none())
```
this only enforces the same behavior for the 2 remaining functions that use `.one()`.

kegl:
OK, looks good to me to merge.

agramfort:
thx @rth

PR #299, opened 2019-12-06
FIX Store training timestamp

rth:
Store the `training` and `sent_to_training` timestamps, when the submission state changes. The columns already exist in the DB but were empty so far.

This also allows to properly display the "time between submission and training" plot in admin dashboad
Screenshot_2019-12-06 Dashboard submissions - RAMP

Before it was showing only zeros.

### DIFF
--- a/ramp-frontend/ramp_frontend/views/admin.py
+++ b/ramp-frontend/ramp_frontend/views/admin.py
@@ -312,7 +312,9 @@ def dashboard_submissions(event_name):
     name_submissions = [sub.name for sub in submissions]
     cumulated_submissions = list(range(1, 1 + len(submissions)))
     training_sec = [
-        (sub.training_timestamp - sub.submission_timestamp).seconds / 60.
+        (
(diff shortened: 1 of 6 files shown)


ramp-frontend/ramp_frontend/views/admin.py
```diff
@@ -312,7 +312,9 @@ def dashboard_submissions(event_name):
     name_submissions = [sub.name for sub in submissions]
     cumulated_submissions = list(range(1, 1 + len(submissions)))
     training_sec = [
-        (sub.training_timestamp - sub.submission_timestamp).seconds / 60.
```
rth: timedelta.seconds shows the number seconds modulo 1 day, and `total_seconds()` should be used to get the total amount of seconds.

kegl:
Ready for MRG, thanks a lot @rth. @agramfort you merge?

agramfort:
thx @rth

PR #301, opened 2019-12-06
ENH Display event status

maikia:
In the list of problems now the checked box is displayed by the event for which the user is signed up for and squared box if (s)he is not.

#### Any other comments?

1. Are those labels clear enough or do we need to have a legend?
2. Other possibilities:
- we might also consider adding a different label in case the event is opened/not for signing up and 
- when the user waits for approval


rth:
Thanks! For UI, could you also please include some screenshots before/after?

maikia:
before:
before

after:
now

kegl:
Looks cool!

agramfort:
looks neat

to me you have 2 orthogonal infos:
- have you signed up
- is the challenge open / public / closed

this tells me that we need 2 visual elements to convey this message

>

kegl:
"is the challenge open / public / closed" - would be cool. Maybe use color codes, explained at the top?

maikia:
That's the current version:

options

Is it not too a bit of overkill?

rth:
I think using [2 separate badges](https://getbootstrap.com/docs/4.0/components/badge/#contextual-variations) for open/closed and signed-up on each line would be clearer.

rth:
Also maybe we can drop "number of",
```html
 - Iris challenge, submissions = 5, participants = 2,
   <span class="badge">Open</span> <span class="badge">Signed-up</span>
```
might be enough?

agramfort:
I like that. +1

agramfort:
these visuals are nice too. Maybe a bit less self explanatory. But that's
already a neat improvement

>

kegl:
Looks great. Some comments:
1. "is open", not "is opened"
2. I never closed an event (as in putting a closing date < now), so the two states to mark are competitive/collaborative rather than open/close. Closed events make the private leaderboard public and I never liked to do that because that somewhat spoils future events.

maikia:
ok, so to sum up:

we want two badges for each event:
signed up/waiting for approval/not signed up
competitive/collaborative 

we assume there is never closing date. what's the point of being able to set closing date then?
can the event be not yet open?

agramfort:
I would say:

signed up/waiting for approval/not signed up
competitive/collaborative/closed

for old challenges I would make them as closed just to limit use of computational ressources if we want. For example running Mars crater submission will break anyway as AWS is not setup for GPUs now.

maikia:
Ok. sounds good. 
Although setting the old challenges as closed would have to be done in the database, ie ramp server. I don't think it makes sense to do it in .html

agramfort:
yes agreed. would this be a change in the DB? If so I would the change. If
the DB does not have the notion of closed challenge than collab or
non-collab is enough

>

maikia:
In the database there is an entry:
events -> closing_timestamp 
this would have to be set to some old date. 
It can be done from the admin level from the webpage

agramfort:
we're good then.

>

maikia:
This is the newest version:

v4 7

what do you think?

kegl:
Nice :)

ramp-frontend/ramp_frontend/views/ramp.py
```diff
@@ -80,10 +80,37 @@ def problems():
...
+        for event in problem.events:
+            # check the state of the event
+            now = datetime.datetime.now()
+            start = event.opening_timestamp
+            start_collab = event.public_opening_timestamp
+            end = event.closing_timestamp
+            if now < start or now >= end:
+                event.state = 'close'
+            elif now >= start and now < start_collab:
+                event.state = 'competetive'
```
agramfort: proper spelling is competitive I think

maikia: Lol, you are right. I will change it now

agramfort:
@rth feel free to merge if you're happy.

thx @maikia

rth:
@agramfort I don't have the commit bit in this repo..

agramfort:
now you can :)

>

ramp-frontend/ramp_frontend/views/ramp.py
```diff
@@ -80,10 +80,37 @@ def problems():
...
+
+    for problem in problems:
+        for event in problem.events:
+            # check the state of the event
+            now = datetime.datetime.now()
+            start = event.opening_timestamp
+            start_collab = event.public_opening_timestamp
+            end = event.closing_timestamp
+            if now < start or now >= end:
+                event.state = 'close'
```
rth: So this is setting an attribute to a SQLAlchemy table object for a field that doesn't exist in the database (and so it has no impact on the database), right? It might be worth mentioning it in a comment.

agramfort:
feel free to merge if it works as expected

thx @rth

>

rth:
Thanks @maikia !

maikia:
> Never-mind, seems to work as expected now. Not sure what changed.

It could have happened if the cache was not cleared after previous sessions

Thanks @rth and @agramfort

Issue #297, opened 2019-12-05
Don't run nbconvert in pytest fixtures


rth:
Currently pytest fixtures that creates the test problem will also run nbconvert on notebooks, which is relatively slow. For instance, running `session_scope_module` fixture from `ramp-database/ramp_database/tools/tests/test_submission.py` produces,
```
[NbConvertApp] Converting notebook iris_starting_kit.ipynb to html
/home/rth/.miniconda3/envs/testenv/lib/python3.7/site-packages/nbconvert/filters/highlight.py:130: UserWarning: IPython lexer unavailable, falling back on Python
  warn("IPython lexer unavailable, falling back on Python")
[NbConvertApp] Writing 361637 bytes to iris_starting_kit.html
[NbConvertApp] Converting notebook boston_housing_starting_kit.ipynb to html
/home/rth/.miniconda3/envs/testenv/lib/python3.7/site-packages/nbconvert/filters/highlight.py:130: UserWarning: IPython lexer unavailable, falling back on Python
  warn("IPython lexer unavailable, falling back on Python")
[NbConvertApp] Writing 456820 bytes to boston_housing_starting_kit.html
```

`nbconvert` HTML generation should be disabled for most tests to help make the test startup faster. 

cc @maikia

PR #303, opened 2019-12-10
ENH add option to create html from notebook using nbconvert

maikia:
#297 

#### Any other comments?
You are right @rth 
it seems that the jupyter notebook / nbconvert is called every time the toy database is created which seems pointless. I don't think this is a problem with fixture though. However, I am not sure why it has been done and that's why I am a bit hesitant to removing it all together. Do you know what is a reason for it @glemaitre @agramfort ?

in ramp_database-testing.py 
create_toy_db() calls
setup_toy_db() calls 
add_problems() calls 
setup_ramp_kit_ramp_data() calls
subprocess.check_output(["jupyter", "nbconvert", "--to", "html",
                             "{}_starting_kit.ipynb".format(problem_name)])

Jupyter jupyter notebook / nbconvert is also tested in: ramp_utils/deploy.py -> deploy_ramp_event


kegl:
Because adding the problem means to render the notebook for the problem landing page.

We could get rid of this by not rendering the notebook on the site, just linking the notbook in the github repo. I'd keep it the way it is if it's not a huge problem, I think it's nice to have the description in the web page.

agramfort:
I agree that having the starting kit embedded in the web page is nice and
running nbconvert allows to make sure it's running without crash

>

ramp-database/ramp_database/testing.py
```diff
@@ -123,7 +123,8 @@ def _delete_line_from_file(f_name, line_to_delete):
         f.truncate()
 
 
-def setup_ramp_kit_ramp_data(ramp_config, problem_name, force=False):
+def setup_ramp_kit_ramp_data(ramp_config, problem_name, force=False,
+                            check_notebook=False):
```
rth: Maybe,
```suggestion
                            convert_notebook=True):
```
and then explicitly set to to false for unit tests? For event deployment I agree that it's nice to have an HTML notebook. Notebook preview on github mostly doesn't work.

glemaitre:
@maikia I think this is fine to have the option which by default does not change the behaviour (as proposed by @rth). However, we are checking that the hml file will exist. So we need to create an empty html file to emulate the conversion. We need a test for this change as well (check that the file is empty).

Do you think that you can finish this PR?

glemaitre:
I will finish this up

glemaitre:
Thanks for the idea @rth. The tests are twice faster.

PR #314, opened 2019-12-15
[WIP] BUG do not load predictions whenever fetching unrelated infos

### LINKED ISSUE #226: Follow-up of #190 - Memory efficient request
The following pattern would keep weakref of  `SubmissionOnCVFold` which will load into memory the prediction arrays.

```python
submissions = [get_submission_by_id(session, sub[o])
               for sub in get_submissions(session, 'iris_test', state='scored')]
for sub in submissions:
    for cv_fold in sub.on_cv_folds:
        ... # for instance get_scores, get_time, etc.
```

If submissions contains submissions then, the predictions arrays will be kept. We allievate the issue with #221 by avoiding to have `submissions` in memory and the arrays for all submissions.

However, there is a better job to be done. Getting the score, time, or memory info should not require to load the array. In facts, these info are stored in `SubmissionScoreOnCVFold`. Therefore, we could make a memory/speed efficient query avoiding to load the arrays:

```python
    submission = get_submission_by_id(session, submission_id)
    results = defaultdict(list)
    for fold_id, cv_fold in enumerate(submission.on_cv_fold):
        results['fold'].append(fold_id)
        for step in ('train', 'valid', 'test'):
            results[step].append(getattr(cv_fold, '{}_time'.format(step)))
    return pd.DataFrame(results).set_index('fold')
```

```python
for step in ('train', 'valid', 'test'):
    times_per_fold = session.query(getattr(SubmissionOnCVFold, '{}_time')).filter(SubmissionOnCVFold.submission_id == submission_id).all()
    for fold_id, times in enumerate(times_per_fold):
        results['fold'].append(fold_id)
        results[step].append(times)
    return pd.DataFrame(results).set_index('fold')
```

glemaitre:
closes #226 

Avoiding to load the predictions when computing the leaderboard.

Let's try to get a regression test.

### DIFF
--- a/ramp-database/ramp_database/model/datatype.py
+++ b/ramp-database/ramp_database/model/datatype.py
@@ -44,4 +44,7 @@ class NumpyType(TypeDecorator):
         array : ndarray
             The NumPy array which has been loaded.
         """
-        return pickle.loads(zlib.decompress(value))
+        print('loading numpy array')
+        xx = pickle.loads(zlib.decompress(value))
+        print(xx)
+        return xx
--- a/ramp-database/ramp_database/tools/submission.py
+++ b/ramp-database/ramp_database/tools/submission.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+from itertools import product
 import datetime
 import logging
 import os
@@ -18,6 +19,7 @@ from ..model import Submission
 from ..model import SubmissionFile
 from ..model import SubmissionFileTypeExtension
 from ..model import SubmissionOnCVFold
+from ..model import SubmissionScoreOnCVFold
 from ..model import SubmissionSimilarity
 from ..model import UserInteraction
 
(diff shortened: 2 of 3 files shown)


agramfort:
@tomMoral this seems better than the joblib cache thing.

glemaitre:
All the `set_*` functions are still loading the predictions. We need to access the attributes directly if we want to avoid loading the predictions.

ramp-database/ramp_database/tools/submission.py
```diff
@@ -626,10 +644,8 @@ def set_predictions(session, submission_id, path_predictions):
     path_predictions : str
         The path where the results files are located.
     """
-    all_cv_folds = (session.query(SubmissionOnCVFold)
-                           .filter_by(submission_id=submission_id)
-                           .all())
-    for fold_id, cv_fold in enumerate(all_cv_folds):
+    sub = get_submission_by_id(session, submission_id)
```
rth: As a side note I'm wondering about the need for some these helper functions. For instance, `get_submission_by_id` will call internally `select_submission_by_id` and then call 
```py
    submission.event.name
    submission.team.name
```
Unless I'm mistaken that's 3 DB queries, for something that could have been just,
```py
sub = (session.query(Submission)
                   .filter(Submission.id == submission_id)
                   .first())
```
it's slightly more verbose, but much more explicit about what is going on..


And actually I find the previous code more explicit without with fewer sqlalchemy magic. Are you saying this is faster or am I missing something?

glemaitre: I agree. I think that we add some refactoring to create some functions for some query but I am unsure of the use indeed. The thing is that, you finish up by writing the query instead of checking that the query was already written. I would be inclined to remove these queries and just be explicit.

ramp-database/ramp_database/model/datatype.py
```diff
@@ -44,4 +44,7 @@ def process_result_value(self, value, dialect):
         array : ndarray
             The NumPy array which has been loaded.
         """
-        return pickle.loads(zlib.decompress(value))
+        print('loading numpy array')
```
glemaitre: Temporary to detect when the predictions are loaded

glemaitre:
FYI so I am sure that we are not loading the predictions when scoring. This is a bit more complex but still readable I think.

I really would like to have a test for that.

PR #320, opened 2019-12-16
BUG show waiting for approval when user sign-up

glemaitre:


ramp-database/ramp_database/tools/frontend.py
```diff
@@ -140,3 +145,27 @@ def is_user_signed_up(session, event_name, user_name):
             (event_team.is_active and event_team.approved)):
         return True
     return False
+
+
+def did_user_signed_up(session, event_name, user_name):
```
rth: Maybe let's call this, `is_user_sign_up_requested` to make it more consistent with the `is_user_signed_up`?

agramfort:
+1 for rename to is_user_sign_up_requested

>

glemaitre:
I will make a PR to correct that. Thanks for the feedback :)

PR #321, opened 2019-12-16
BUG sort cv folds ordered during queries

### LINKED ISSUE #292: scoring bug
Fold predictions and scores are scrambled. We assume here 
https://github.com/paris-saclay-cds/ramp-board/blob/master/ramp-database/ramp_database/tools/submission.py#L632
that folds are always enumerated in the same order but they seem not to be. It is also possible that the problem is when the folds are created here: https://github.com/paris-saclay-cds/ramp-board/blob/master/ramp-database/ramp_database/tools/submission.py#L95

In any case, a lot of times the scores are assigned to wrong folds. It seems to be not affecting the leaderboard scores, only the combined score, but I am not sure. If they do, it's a quite serious bug.

glemaitre:
closes #292 

The queries on cv folds do not ensure to get always the fold in the exact same order. Therefore, we always sort the fold based on their creation ID.

### DIFF
--- a/ramp-database/ramp_database/model/submission.py
+++ b/ramp-database/ramp_database/model/submission.py
@@ -408,6 +408,7 @@ class Submission(Model):
             all_cv_folds = (session.query(SubmissionOnCVFold)
                                    .filter_by(submission_id=self.id)
                                    .all())
+            all_cv_folds = sorted(all_cv_folds, key=lambda x: x.id)
         for submission_on_cv_fold in all_cv_folds:
             submission_on_cv_fold.state = state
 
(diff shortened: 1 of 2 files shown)


ramp-database/ramp_database/model/submission.py
```diff
@@ -408,6 +408,7 @@ def set_state(self, state, session=None):
             all_cv_folds = (session.query(SubmissionOnCVFold)
                                    .filter_by(submission_id=self.id)
                                    .all())
+            all_cv_folds = sorted(all_cv_folds, key=lambda x: x.id)
```
rth: Thanks! I think,
```py
  .filter_by(submission_id=self.id)
  .order_by(SubmissionOnCVFold.id)
  .all()
```
would be a more efficient way of doing this as sorting is then done at the DB level.

So there shouldn't be issues with backward compatibility here, right?

PR #324, opened 2019-12-17
ENH friendlier side-bar in frontend

### LINKED ISSUE #271: approve users link not just from event page
it should be possible to go to approve users page from any page when you are admin. Currently you need to be in an event page

maikia:
#### Reference Issue

closes #271 

More user-friendly side bar which does not change depending on which page is the logged-in user.

1) the links to the event appear but they cannot be clicked if the user did not select the event. Once the event is selected links to the event-related pages are active

2) for the user which was not yet approved for an event by the administrator some of the links of the event will stay inactive until the user is approved

3) status of the user was not passed to ramp_description.html as it was not needed until now. Now the status of the user is also passed to ramp_description.html

Some examples from before and after (logged in as admin):

1. Before:

- view from the admin user viewing event for which he is not approved

before_event

- view from the admin user not viewing any event

before_non-event

2. After:

- view from the admin user viewing event for which he is not approved

after_event_not_approved

- view from the admin user viewing event for which he is approved

after_event

- view from the admin user not viewing any event

after_non-event

### DIFF
--- a/ramp-frontend/ramp_frontend/static/css/style.css
+++ b/ramp-frontend/ramp_frontend/static/css/style.css
@@ -273,11 +273,6 @@ a:hover {
           line-height: 44px;
           vertical-align: middle;
           padding: 0em 1.2em; }
-        .app-container .content-container .side-menu .navbar-nav .inactiveLink li.dropdown ul li {
-          pointer-events: none;
-          cursor: not-allowed;
-          color: gray
-         }
         .app-container .content-container .side-menu .navbar-nav li.dropdown > a:after {
           content: "";
           position: absolute;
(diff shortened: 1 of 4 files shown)


kegl:
LGTM, just a small remark: it's strange that the site admin can see only the private leaderboards for the events she's not signed up for. I would make the public leaderboards also accessible.

kegl:
My other remark is more general: I would only do frontend updates if there are upcoming RAMPs in the near future where these new features are live-tested.

glemaitre:
I have a RAMP for DSSP in february.

agramfort:
Thx @maikia

PR #325, opened 2019-12-17
ENH disable submit button if closed

maikia:
#### Reference Issue
Right now submission is always possible, no matter if the event is open or not. 
This PR checks if the event is in one of the states:
not yet opened, open, no longer opened

then it displays appropriate message in the sandbox
and if it is not open the submission button becomes disabled:

Couple of examples with messages and disabled buttons:

event open:

enabled

event closed:

disabled


ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -385,6 +386,34 @@ def test_ask_for_event_mail(client_session):
...
+    sign_up_team(session, 'iris_test_4event', 'test_user')
+
+    with login_scope(client, 'test_user', 'test') as client:
+        event_info = {
+            'button': 'Submit'
+        }
+        rv = client.get('http://localhost/events/iris_test_4event/sandbox',
+                        data=event_info)
+        assert rv.status_code == 200
+        if expected == b'event-close':
```
glemaitre: Put a small comment explaining that the button should be disable and it will appear in the html data. 

The reason is that the test is a bit loose since more element could be disabled but this is enough here. However in case of a change in the future it will be difficult to remember what is disabled at a first glance.

ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -385,6 +386,34 @@ def test_ask_for_event_mail(client_session):
                         in outbox[0].body)
 
 
+@pytest.mark.parametrize(
+    "opening_date,public_date,closing_date,expected", testtimestamps
+)
+def test_submit_button_enabled(client_session, makedrop_event,
```
glemaitre:
```suggestion
def test_submit_button_enabled_disabled(client_session, makedrop_event,
```

ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -385,6 +386,34 @@ def test_ask_for_event_mail(client_session):
                         in outbox[0].body)
 
 
+@pytest.mark.parametrize(
+    "opening_date,public_date,closing_date,expected", testtimestamps
```
glemaitre:
```suggestion
    "opening_date, public_date, closing_date, expected", testtimestamps
```

ramp-frontend/ramp_frontend/views/ramp.py
```diff
@@ -319,6 +319,25 @@ def sandbox(event_name):
...
+    now = datetime.datetime.now()
+    start = event.opening_timestamp
+    end = event.closing_timestamp
+
+    event_status = {"msg": "",
+                    "state": "not_yet"}
+    start_str = start.strftime("%d of %B %Y at %H:%M")
+    end_str = end.strftime("%d of %B %Y, %H:%M")
+    if now < start:
+        event_status["msg"] = f"Event submissions will open on the {start_str}"
```
glemaitre: Are we supporting f string. Check that we are supporting at least python 3.6 I think.

maikia: f string works from python 3.6 but I changed it to be sure
Otherwise I think it's ready to go
what do you say @glemaitre ?

glemaitre:
Thanks @maikia

maikia:
Thanks @glemaitre @agramfort

Issue #270, opened 2019-11-06
remove users from frontend


agramfort:
it should be possible to remove to remove users when you are admin from the frontend.

PR #326, opened 2019-12-17
ENH allow to delete a user from database

glemaitre:
#### Reference Issue

partially addressing #270

### DIFF
--- a/ramp-database/ramp_database/tools/user.py
+++ b/ramp-database/ramp_database/tools/user.py
@@ -104,6 +104,21 @@ def add_user(session, name, password, lastname, firstname, email,
     return user
 
 
+def delete_user(session, name):
+    """Delete a user from the database.
+
+    Parameters
+    ----------
+    session : :class:`sqlalchemy.orm.Session`
+        The session to directly perform the operation on the database.
+    name : str
+        The name of the user.
+    """
+    user = session.query(User).filter(User.name == name).one()
+    session.delete(user)
+    session.commit()
+
+
 def make_user_admin(session, name):
     """Make a user a RAMP admin.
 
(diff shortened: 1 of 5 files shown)


ramp-database/ramp_database/tools/user.py
```diff
@@ -104,6 +104,21 @@ def add_user(session, name, password, lastname, firstname, email,
...
+
+    Parameters
+    ----------
+    session : :class:`sqlalchemy.orm.Session`
+        The session to directly perform the operation on the database.
+    name : str
+        The name of the user.
+    """
+    user = session.query(User).filter(User.name == name).one()
+    session.delete(user)
```
agramfort: Should we delete also teams he belonged to and the submissions ?

glemaitre: `session.delete` will make a cascade delete where there is a foreign key. Team should have a foreign key there. What could be added is that it really happen in the team table.

glemaitre: I pushed a quick additional check: https://github.com/paris-saclay-cds/ramp-board/commit/c73e2fa3d7f6bf15a88e7b72dc45aeb8e2cac69c

agramfort:
great

corresponding submissions are deleted too?

when you delete a submission you remove files from disk?
bagged score should then technically be recomputed if this submission was
present
but it's for another iteration. Just raising the issue

>

glemaitre:
> corresponding submissions are deleted too?

If they have a foreign key link to the user, they will be.

> when you delete a submission you remove files from disk?

At least from the database

>bagged score should then technically be recomputed if this submission was
present but it's for another iteration. Just raising the issue

True

PR #329, opened 2019-12-18
ENH multi-extension&symlink non-editable for files + cache score computations

tomMoral:
This PR back-port the fixes I had to make to make the server run:
* Use a make develop
* Allow for multi-extension
* Cache score computations

I also included a fix to allow access to code in collab phase.


Makefile
```diff
@@ -15,6 +15,12 @@ install:
 	cd ramp-engine && pip install . && cd ..
 	cd ramp-utils && pip install . && cd ..
 
+develop:
+	cd ramp-frontend && pip install -e . && cd ..
+	cd ramp-database && pip install -e . && cd ..
+	cd ramp-engine && pip install -e . && cd ..
+	cd ramp-utils && pip install -e . && cd ..
```
rth: Thant's identical to `inplace` below..

ramp-database/ramp_database/tools/submission.py
```diff
@@ -68,7 +69,7 @@ def add_submission(session, event_name, team_name, submission_name,
     submission = (session.query(Submission)
                          .filter(Submission.name == submission_name)
                          .filter(Submission.event_team == event_team)
-                         .one_or_none())
+                         .first())
```
rth: I would rather keep `one_or_none` then explicitly handle the None case. Otherwise .first() would return an empty list, if none match, right?

Actually, `None` is already handled below, why is this necessary then?

ramp-database/ramp_database/tools/submission.py
```diff
@@ -415,18 +425,26 @@ def get_time(session, submission_id):
     computation_time : pd.DataFrame
         A pandas dataframe containing the computation time of each fold.
     """
-    results = defaultdict(list)
-    all_cv_folds = (session.query(SubmissionOnCVFold)
+    cache_filename = os.path.join(
+            CACHE_DIR, 'times_submission_{:09}.pkl.zip'
+            .format(submission_id))
+    if os.path.exists(cache_filename):
+        times = pd.read_pickle(cache_filename)
```
rth: Not really sure what is going on here, could you explain? (and also if some of this wasn't addressed in some of the other PRs).

ramp-database/ramp_database/tools/submission.py
```diff
@@ -444,21 +462,28 @@ def get_scores(session, submission_id):
     scores : pd.DataFrame
         A pandas dataframe containing the scores of each fold.
     """
-    results = defaultdict(list)
-    index = []
-    all_cv_folds = (session.query(SubmissionOnCVFold)
+    cache_filename = os.path.join(
+            CACHE_DIR, 'scores_submission_{:09}.pkl.zip'
```
rth: Same as above

ramp-frontend/ramp_frontend/views/ramp.py
```diff
@@ -823,14 +823,20 @@ def view_model(submission_hash, f_name):
                         submission, filename)
             )
 
+            workflow_element = WorkflowElement.query.filter_by(
+                name=filename.split('.')[0], workflow=event.workflow).one()
```
rth: Maybe define a separate helper function to get root ° extension from a path, and use it consistently? `os.path.splitext` is not doing what we need?

agramfort:
@tomMoral can you see @rth comment's before we merge this?

ramp-database/ramp_database/tools/submission.py
```diff
@@ -32,6 +32,7 @@
 from ._query import select_submission_file_type_by_name
 from ._query import select_team_by_name
 
+CACHE_DIR = '/home/ramp/ramp_deployment/cache/'
```
kegl: I think this should not be hardcoded here, rather handled using config.

glemaitre: We used sqlalchemy to differ the loading. We don't need caching anymore.

PR #332, opened 2019-12-18
[MRG] Update index page

### LINKED ISSUE #331: Update frontend index page
Update the home page (`index.html`), especially for the new OVH server:

As discussed with @glemaitre and @agramfort;
- remove CDS/Paris Saclay specific things. If ramp board is intended to be a package used by others to setup their own server, the home page should be more general instead of looking like it is only for CDS
- make it clear that ramp board is both a bundle of packages that allow you to setup your own server and a platform for ML challenges

Is this an accurate summary of our discussion? What do you think @kegl?

lucyleeow:
closes #331


agramfort:
can you share screen shots?

>

lucyleeow:
Still fiddling with the wording. Suggestions welcome. @agramfort 

Screenshot_2019-12-18 RAMP collaborative data science challenges

lucyleeow:
We should tweak the colour scheme (@glemaitre):

image

lucyleeow:
@agramfort I made your face bigger! :smile:

glemaitre:
I like the contrast of the first part. We just need a bit more space between "Python" and the button.

I would come back on my idea to have 2 background (white and light-green) it does not look that good. Probably a hline would be enough to separate the different part (similarly to dask werbsite)

glemaitre:
We probably need a font that has more spacing between letters naturally. I find it very compact.
And just missing the RAMP logo in the bar :)

Apart from that, I am really happy with the general look. We are only at the nitpicks level now.

glemaitre:
Thanks @lucyleeow for the effort

lucyleeow:
Thank you for the excellent button suggestion @glemaitre - it looks much better now.

Should I add horizontal lines after each section?

image

lucyleeow:
FYI - I did not include any of the JS files so you can start afresh.

agramfort:
anything specific you want me to edit before merging?

lucyleeow:
@agramfort you see horizontal line between the RAMP data challenges and RAMP packages? Should I add the same horizontal line between RAMP packages and RAMP in action sections & between RAMP in action and Developed by sections?

lucyleeow:
This is what it looks like with the extra lines - now that I look at it I think it's better like this:

image

lucyleeow:
Yes that was the plan! I will merge now, thanks!

kegl:
I like it. I would have kept the photo though.

PR #333, opened 2019-12-18
BUG ensure to remove all attached keys

glemaitre:
Adding a delete cascade for `User` which was not there by default. Updated the tests as well.


glemaitre:
Apparently the cascade does not work

PR #336, opened 2019-12-19
MNT remove some typo from emails sent user/admin

glemaitre:
Fix some typos from the emails sent to the admins and users.


glemaitre:
We should add test by:

* entering in a competition phase
* check that we have the button put-in /pull-out
* check that we can trigger pull-in / pull-out

agramfort:
thx @glemaitre

PR #339, opened 2020-01-07
ENH add option to delete an event

maikia:
For now it adds the test to make sure that when an event is deleted its reference is deleted from all the tables.

I think it would be of interest to add a button enabling an event admin to delete an existing event.
This could either only delete the event from the database or also remove event directories from the disk.
What do you think @glemaitre @agramfort ?


agramfort:
Removing an event should delete all related submissions 

Agreed to have a button in event admin page

maikia:
I added the remove event button. It will remove the event from the database but not from the disk. 

Before:
before

After:
after2

after1

What do you think @agramfort @glemaitre ?

agramfort:
any reason not to remove from the disk?

>

glemaitre:
> any reason not to remove from the disk?

It could be dangerous to do that. If for instance, you just want to move the event folder and just recreate the event in the database, it will delete the file and you are screwed up :)

Since this is not difficult to remove manually the file or even you can `--force` while deploying to overwrite an event, this should be safer to go this way.

ramp-database/ramp_database/tools/tests/test_event.py
```diff
@@ -263,6 +272,68 @@ def test_check_event(session_scope_function):
...
+    event_test = get_event(session_scope_function, None)
+    assert len(event_test) == 0
+    assert not get_event_team_by_name(session_scope_function,
+                                      event_name, username)
+    assert not get_event_admin(session_scope_function, event_name, username)
+    event_score_types = get_score_type_by_event(session_scope_function, event)
+    assert len(event_score_types) == 0
+    event_cv_fold = get_cv_fold_by_event(session_scope_function, event)
+    assert len(event_cv_fold) == 0
+    with pytest.raises(AttributeError):
```
glemaitre: I would instead my the query finishing with all() and check that we have an empty list

ramp-database/ramp_database/tools/tests/test_event.py
```diff
@@ -263,6 +272,68 @@ def test_check_event(session_scope_function):
...
+    # make sure event and all the connections were deleted
+    event_test = get_event(session_scope_function, None)
+    assert len(event_test) == 0
+    assert not get_event_team_by_name(session_scope_function,
+                                      event_name, username)
+    assert not get_event_admin(session_scope_function, event_name, username)
+    event_score_types = get_score_type_by_event(session_scope_function, event)
+    assert len(event_score_types) == 0
+    event_cv_fold = get_cv_fold_by_event(session_scope_function, event)
+    assert len(event_cv_fold) == 0
```
glemaitre:
```suggestion
    assert not event_cv_fold
```

glemaitre: This usually more python to check list emptiness

ramp-database/ramp_database/tools/event.py
```diff
@@ -433,6 +433,48 @@ def get_event(session, event_name):
...
+    ----------
+    session : :class:`sqlalchemy.orm.Session`
+        The session to directly perform the operation on the database.
+    event_name : str or None
+        The event class to query
+
+    Returns
+    -------
+    score type : :class:`ramp_database.model.ScoreType` or \
+    list of :list of :class:`ramp_database.model.ScoreType`
```
glemaitre:
```suggestion
    list of list of :class:`ramp_database.model.ScoreType`
```

ramp-database/ramp_database/tools/event.py
```diff
@@ -433,6 +433,48 @@ def get_event(session, event_name):
...
 
+def get_cv_fold_by_event(session, event):
+    """Get ScoreType from the database
+
+    Parameters
+    ----------
+    session : :class:`sqlalchemy.orm.Session`
+        The session to directly perform the operation on the database.
+    event_name : str or None
+        The event class to query
```
glemaitre:
```suggestion
        The event class to query.
```

ramp-database/ramp_database/tools/event.py
```diff
@@ -433,6 +433,48 @@ def get_event(session, event_name):
...
+    -------
+    cv fold : : list of all cv folds of this event
+    """
+    return (session.query(CVFold)
+                   .filter(EventScoreType.event_id ==
+                           event.id)
+                   .all())
+
+
+def get_score_type_by_event(session, event):
```
glemaitre: We have an issue open where we tout with Roman to revert creating these functions and instead create the query directly. It is not saving so much line of code and this is explicit enough to make the query

ramp-database/ramp_database/tools/tests/test_event.py
maikia: coming back to your other comment about not writing the get_ queries: would you write it the the test_event.py itself?

glemaitre:
OK so since this is not easy we can first make that we can delete the event from DB
Then we can always add another elegant solution for the file.

ramp-frontend/ramp_frontend/templates/update_event.html
```diff
@@ -133,7 +133,34 @@ <h4 class="modal-title" id="myModalLabel">{{ category }}</h4>
                            <span style="color: red;">[{{ error }}]</span>
                        {% endfor %}
                    </div>
-                   <input type="submit" value="Update" class="ui submit button">
+                   <button type="submit" name="submit_button" value="Update!" class="ui submit button">Update</button>
```
glemaitre: So we only need to remove this part of the html

ramp-frontend/ramp_frontend/templates/update_event.html
```diff
@@ -133,7 +133,34 @@ <h4 class="modal-title" id="myModalLabel">{{ category }}</h4>
...
+                   <!-- Button trigger modal -->
+                    <button type="button" class="ui remove button" data-toggle="modal" data-target="#exampleModal">
+                        Remove event
+                    </button>
+                    
+                    <!-- Modal -->
+                    <div class="modal fade" id="exampleModal" tabindex="-1" role="dialog" aria-labelledby="exampleModalLabel" aria-hidden="true">
```
glemaitre: But in the end it was really useful for the sandbox submission :)

ramp-frontend/ramp_frontend/tests/test_admin.py
```diff
@@ -15,6 +15,7 @@
 from ramp_database.utils import setup_db
 from ramp_database.utils import session_scope
 
+from ramp_database.tools.event import add_event
```
glemaitre: to be removed as well

ramp-frontend/ramp_frontend/views/admin.py
```diff
@@ -221,43 +222,48 @@ def update_event(event_name):
...
-            # #     if len(message) > 0:
-            # #         message += ' and '
-            # #     message += 'email is already in use'
-            # except NoResultFound:
-            #     pass
-            if message:
-                e = NameClashError(message)
-            flash('{}'.format(e), category='Update event error')
-            return redirect(url_for('update_event', event_name=event.name))
+        if request.form["submit_button"] == "Confirm!":
```
glemaitre: remove as well

ramp-frontend/ramp_frontend/views/admin.py
```diff
@@ -221,43 +222,48 @@ def update_event(event_name):
...
+            delete_event(db.session, event_name)
```
glemaitre: So we will need to add a CLI interface for this function, add the documentation.
Regarding the CLI option, I would add a flag option to delete the file on the disk (you can see how to do that here: https://github.com/paris-saclay-cds/ramp-workflow/blob/master/rampwf/utils/cli/testing.py#L41)

basically one should run `ramp database delete-event --name xxx --from-disk` (where `--from-disk` would be the flag)

maikia:
@glemaitre can you have a look? If you agree I will add the docs

now the user needs to pass the path to the events config.yml and event_name which might be annoying. event_name could easily be read from the config.yml 
(unless it changes and it's different in the config.yml file and in the db, is that possible?). 
However, if those two params are kept all the remains will be removed (ie from the db even if there is no more config.yml on the disk or the other way around).

ramp-frontend/ramp_frontend/templates/update_event.html
```diff
@@ -147,5 +147,10 @@ <h4 class="modal-title" id="myModalLabel">{{ category }}</h4>
 $(function () {
    $('#formerror').modal('toggle')
 });
+$(document).keypress(function(e) {
```
glemaitre: What it is useful for?

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,52 @@ def add_submission(config, event, team, submission, path):
...
+                      will be removed from the database and from the disk;
+                      if set to False they will only be removed from
+                      the database""")
+def delete_event(config, config_event, event_name, from_disk):
+    """Delete event."""
+    config = read_config(config)
+
+    # check if the event_config dir exists
+    if not os.path.isfile(config_event) and from_disk:
+        print('cannot find '+config_event)
```
glemaitre: I don't think that we should use print when using click.

It should be `click.echo()`: https://click.palletsprojects.com/en/7.x/utils/
We can try to use `err=True` as well.

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,52 @@ def add_submission(config, event, team, submission, path):
...
+        from_disk = False
```
glemaitre: Actually we should be raising an error to not go further. The echo something like

```python
click.echo(
    "The event config file is not existing. This file is required for"
    "deleting files from the disks. Either remove the flag --from-disk"
    "or give the correct path to the event config file.",
    err=True
)
```

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,52 @@ def add_submission(config, event, team, submission, path):
...
+        db_event = event_module.get_event(session, event_name)
+        if db_event:
+            event_module.delete_event(session, event_name)
+            if from_disk:
+                # remove submissions dir from disk if it exists
+                path_to_submissions = db_event.path_ramp_submissions
+                if os.path.isfile(path_to_submissions):
+                    shutil.rmtree(path_to_submissions)
+                    print('removed directory '+path_to_submissions)
+        if db_event:
```
glemaitre: uhm you already have the same `if` statement above. It seems this not going to work. I would probably do this structure instead:

```python
db_event = event_module.get_event(session, event_name)
if not db_event:
    echo.click("The event {} is not in the database. Aborting!!!"
               .format(event_name), err=True)
# code to delete
```

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,52 @@ def add_submission(config, event, team, submission, path):
...
+        # is event in the database?
+        db_event = event_module.get_event(session, event_name)
+        if db_event:
+            event_module.delete_event(session, event_name)
+            if from_disk:
+                # remove submissions dir from disk if it exists
+                path_to_submissions = db_event.path_ramp_submissions
+                if os.path.isfile(path_to_submissions):
+                    shutil.rmtree(path_to_submissions)
+                    print('removed directory '+path_to_submissions)
```
glemaitre:
```suggestion
                    click.echo("Remove the submissions: {}"
                               .format(path_to_submissions))
```

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,52 @@ def add_submission(config, event, team, submission, path):
...
+    # remove event from the database
+    with session_scope(config['sqlalchemy']) as session:
+        # is event in the database?
+        db_event = event_module.get_event(session, event_name)
+        if db_event:
+            event_module.delete_event(session, event_name)
+            if from_disk:
+                # remove submissions dir from disk if it exists
+                path_to_submissions = db_event.path_ramp_submissions
+                if os.path.isfile(path_to_submissions):
```
glemaitre: This is weird here. If `path_to_submissions` is a directory, then we should use something else than `isfile`. If it is a list of submissions, I would expect that we loop other the file one by one?

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,52 @@ def add_submission(config, event, team, submission, path):
...
+                    shutil.rmtree(path_to_submissions)
+                    print('removed directory '+path_to_submissions)
+        if db_event:
+            event_module.delete_event(session, event_name)
+        else:
+            print('!!'+event_name+' was not found in the database')
+    if from_disk:
```
glemaitre: This should be moved within the above statement.

Basically we want to only remove if it is present in the database.

glemaitre: Maybe you are thinking that we might forget to remove the file and then we would like to do that but there is not event in the database anymore. In this case, I think that we can add a `--force` flag which will remove all files on disk even if `db_event` is empty. Does this make sense?

agramfort:
tell us when it's good to go @maikia 

thx

maikia:
> tell us when it's good to go @maikia 
> thx

I would prefer if @glemaitre could review before we merge this one

agramfort:
ok !

>

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,80 @@ def add_submission(config, event, team, submission, path):
...
+                      (if it exists) even if it is not found
+                      in the database.
+                      This option must be used with --from-disk
+                      option to work""")
+def delete_event(config, config_event, from_disk, force):
+    """Delete event."""
+    config = read_config(config)
+
+    # check if the event_config dir exists
+    if not os.path.isfile(config_event):
```
glemaitre: I am not sure that we need this. If we provide something that it is not correct it will fail, isn't it?

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,80 @@ def add_submission(config, event, team, submission, path):
...
+            db_event = event_module.get_event(session, event_name)
+            if db_event:
+                event_module.delete_event(session, event_name)
+
+                if from_disk:
+                    # remove submissions dir from disk if it exists
+                    path_to_submissions = db_event.path_ramp_submissions
+                    if os.path.exists(path_to_submissions):
+                        shutil.rmtree(path_to_submissions)
+                        click.echo("removed directory " +
```
glemaitre:
```suggestion
                        click.echo("Removed directory {}"
```

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,80 @@ def add_submission(config, event, team, submission, path):
...
+                                   "{}".format(path_to_submissions))
```
glemaitre:
```suggestion
                                   .format(path_to_submissions))
```

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,80 @@ def add_submission(config, event, team, submission, path):
...
+                    else:
+                        click.secho("No such submissions directory: " +
```
glemaitre:
```suggestion
                        click.secho("No such submissions directory: {}"
```

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,80 @@ def add_submission(config, event, team, submission, path):
...
+                                    "{}".format(path_to_submissions),
```
glemaitre:
```suggestion
                                    .format(path_to_submissions),
```

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,80 @@ def add_submission(config, event, team, submission, path):
...
+                                    fg='red', err=True)
+            else:
+                click.secho("No such event in the " +
```
glemaitre:
```suggestion
                click.secho("No such event in the "
```

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,80 @@ def add_submission(config, event, team, submission, path):
...
+                            "database: {}".format(event_name),
+                            fg='red', err=True)
+                if from_disk and not force:
+                    click.echo("please use options --force --from_disk " +
```
glemaitre:
```suggestion
                    click.echo("Please use options --force --from_disk "
```

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,80 @@ def add_submission(config, event, team, submission, path):
...
+                               "if you would like to remove from the " +
```
glemaitre:
```suggestion
                               "if you would like to remove from the "
```

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,80 @@ def add_submission(config, event, team, submission, path):
...
+                    path_to_submissions = db_event.path_ramp_submissions
+                    if os.path.exists(path_to_submissions):
+                        shutil.rmtree(path_to_submissions)
+                        click.echo("removed directory " +
+                                   "{}".format(path_to_submissions))
+                    else:
+                        click.secho("No such submissions directory: " +
+                                    "{}".format(path_to_submissions),
+                                    fg='red', err=True)
+            else:
```
glemaitre: we need before this else the
```python
elif from_disk and force:
    # forcing to remove from the disk.
```
In this case we don't get any message that the event is not in the database.

ramp-database/ramp_database/cli.py
```diff
@@ -189,6 +191,80 @@ def add_submission(config, event, team, submission, path):
...
+                if from_disk and not force:
+                    click.echo("please use options --force --from_disk " +
+                               "if you would like to remove from the " +
+                               "disk events directory: {}".format(event_dir))
+            if event_dir:
+                if (db_event or force) and from_disk:
+                    # remove this event from the disk
+                    shutil.rmtree(event_dir)
+                    click.echo("removed directory: {}".format(event_dir))
+                elif db_event:
```
glemaitre: This will not be useful if we have integrated in the previous `if` branching

agramfort:
Shall I merge @glemaitre ?

glemaitre:
yes

glemaitre:
Ups no.

glemaitre:
No it is fine. I got scared with an not up-to-date tab on my firefox.

PR #340, opened 2020-01-08
removed the dots from the list of events

maikia:
Removed the dots from the list in the list of events

before:
before

after:
after


lucyleeow:
LGTM, merging.

PR #341, opened 2020-01-08
[MRG] Frontend: teaching page

lucyleeow:
Includes information from the old about page and the teaching page.

image

Old page:

image

cc @agramfort @kegl

### DIFF
--- a/ramp-frontend/ramp_frontend/templates/teaching.html
+++ b/ramp-frontend/ramp_frontend/templates/teaching.html
@@ -39,11 +39,11 @@
 
             <p>
                 RAMP challenges begin with an interesting supervised prediction
-                problem for which there is data. Domain experts then work with
-                experienced data scientists to clean and curate the
-                data and formalize the problem. This process can take two weeks to
-                six months, and results in a 'starting kit', which typically
-                consists of:
+                problem for which there is data. Domain experts teaching
+                clean and curate the data and formalize the problem, working
+                with experienced data scientists if required. This process can
+                take two weeks to six months, and results in a 'starting kit',
+                which typically consists of:
 
                 <ul>
                     <li>
@@ -89,10 +89,9 @@
             <p>
                 <h1 style='font-weight: bold;'>Using RAMP for teaching</h1>
                 If you are interested in using the platform for teaching,
-                We can open a new event for any <a href="/problems">
-                RAMP challenge</a>. You will have full control over the users
-                who can sign up for your event and the timeline the event
-                (start, moving from closed to open phase and end).
+                we can open a new event for any <a href="/problems">
+                RAMP challenge</a>. You will have full control over the timeline
+                of the event (start, moving from closed to open phase and end).
             </p>
 
             <p>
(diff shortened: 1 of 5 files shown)


ramp-frontend/ramp_frontend/templates/teaching.html
```diff
@@ -27,27 +27,107 @@ <h4 class="modal-title" id="myModalLabel">{{ category }}</h4>
...
+
+            <br>
+
+            <p>
+                <h1 style='font-weight: bold;'>Using RAMP for teaching</h1>
+                If you are interested in using the platform for teaching,
+                We can open a new event for any <a href="/problems">
+                RAMP challenge</a>. You will have full control over the users
+                who can sign up for your event and the timeline the event
+                (start, moving from closed to open phase and end).
```
agramfort: this assumes that teachers are admins. You cannot be admin of only one event for now.

ramp-frontend/ramp_frontend/templates/teaching.html
```diff
@@ -27,27 +27,107 @@ <h4 class="modal-title" id="myModalLabel">{{ category }}</h4>
...
+            </p>
+
+            <p>
+                At the end of your challenge we will give you a csv file with all
+                the performance metrics of each student.
```
agramfort: it should be able to download this from the admin page from the event. It's very manual so far

ramp-frontend/ramp_frontend/templates/teaching.html
```diff
@@ -27,27 +27,107 @@ <h4 class="modal-title" id="myModalLabel">{{ category }}</h4>
...
+                thousands of models that are already in the platform.
+            </p>
+
+            <br>
+
+            <p>
+                <h1 style='font-weight: bold;'>Contact us</h1>
+
+                To use RAMP for teaching or to work with us to develop a new
+                RAMP challenge, contact us at <a href="mailto:admin@ramp.studio"><strong>admin@ramp.studio</strong></a>.
```
agramfort: I would also that RAMP code is open source so you can easily set up your own server. That's the easiest way to make your own challenge. It does not require our time

ramp-frontend/ramp_frontend/templates/teaching.html
lucyleeow: So just remove the sign up part?

ramp-frontend/ramp_frontend/templates/teaching.html
lucyleeow: I don't see this option. Should it be under 'dashboard admin' or 'update event'? @glemaitre @maikia do you know if this option exists?

ramp-frontend/ramp_frontend/templates/teaching.html
agramfort: yes. Don't say that they will be able to approve their users

ramp-frontend/ramp_frontend/templates/teaching.html
agramfort: yes it is not available yet but if we say this it should be added

lucyleeow:
image

lucyleeow:
Good to merge @agramfort!

agramfort:
thx @lucyleeow

PR #342, opened 2020-01-09
[MRG] Frontend remove button from index

lucyleeow:
With the update to the description/history page and teach/your challenges page we can get rid of the methodology button (the 'your challenges' button at the top links to the same info)

New button appearance:
image


agramfort:
website updated :)

PR #343, opened 2020-01-10
Improve classify challenge

maikia:
1. It makes shorter/clearer the message about viewing problems through classify.
2. If there are no problems in classify category it displays the message that there are no problems in this category
3. It adds not classified heading for problems which are not classified to any category

**- before:**
message on the problems list:
before1

classified page:
before2

**- after:**
message on the problems list:
after1

classified page:
after2

### DIFF
--- a/ramp-frontend/ramp_frontend/templates/data_domains.html
+++ b/ramp-frontend/ramp_frontend/templates/data_domains.html
@@ -61,6 +61,19 @@
 {% endfor %}
 </ul>
 
+<div class="page-title">
+    <span class="title">Not classified</span>
+</div>
+
+<ul>
+{% for problem in problems %}
+    {% if problem.name not in keywords.problems %}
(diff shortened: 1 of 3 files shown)


lucyleeow:
It's failing because there are some hard coded tests that check for specific events in data domain. This test doesn't seem very robust. Any suggestions @agramfort @glemaitre

glemaitre:
If I remember correctly, we deploy Boston and we check that the domains related to Boston are in the webpage. So this is hardcoded but because of what we deploy.

glemaitre:
So in case, you change something regarding the toy dataset, you can adapt the test accordingly even if it is hard-coded value.

agramfort:
thx @maikia 

website updated

maikia:
thanks @agramfort @lucyleeow @glemaitre

PR #344, opened 2020-01-10
[MRG] Frontend make fonts smaller

lucyleeow:
Should fix the font issues - some of this is due to a different font being used (Source Sans Pro instead of Roberto Condensed - I think this is good for consistency and Roberto was too condensed and not good for accessibility)


lucyleeow:
@agramfort if you are happy can you merge and update the server?

agramfort:
thx @lucyleeow

Issue #330, opened 2019-12-18
BUG event renamed when updated from browser


tomMoral:
When using the `update event` tab from the browser, modifying the event will modify its name.

maikia:
Thanks @tomMoral 
Can you tell me what exactly happened? I don't seem to be able to reproduce it.

PR #345, opened 2020-01-13
BUG ensures event_name proceeded by the problem_name

maikia:
#### Reference Issue

Fixes: #330

The event_name could have been freely chosen when the event was created at the terminal level.
However, when updated from the frontend the event was ensured to have the problem_name as the prefix which sometimes was leading to unexpected change of name of the event

It now checks if the event_name has the problem_name as the prefix set in the config.yml at the creation of the event. if not it throws an error

@tomMoral @agramfort @lucyleeow


agramfort:
good to go?

doc/create_ramp_event.rst
```diff
@@ -42,7 +42,9 @@ This config file should look like::
         conda_env: ramp-iris
 
 .. note::
-    In the previous configuration example, we are using a conda worker. You can
+    - <event_name must always be preceeded with the '<problem_name>_' as in
```
tomMoral: `<event_name` -> Typo?

ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -151,24 +152,47 @@ def test_problem(client_session):
...
+    client, session = client_session
+    if not correct:
+        with pytest.raises(ValueError) as e:
+            add_event(session, 'iris', event_name, 'new_event', 'starting_kit',
+                      '/tmp/databoard_test/submissions', is_public=True)
+        assert "<event_name> must start with" in str(e.value)
+    else:
+        assert add_event(session, 'iris', event_name, 'new_event',
+                         'starting_kit', '/tmp/databoard_test/submissions',
+                         is_public=True)
```
tomMoral: It is great to have a test!

tomMoral:
Fine for me!

agramfort:
Thx @maikia

glemaitre:
Thanks @maikia

maikia:
thanks @agramfort @glemaitre @tomMoral

PR #349, opened 2020-01-14
MNT remove the predictions from the database

### LINKED ISSUE #269: do not store the predictions and the fold indices in the database
storing the predictions in the database makes it huge and we store it on disk anyway in the submission folders. Let's not store it twice and let's keep it out of the DB to make it lighter.

and apparently the same applies to cv indices.

glemaitre:
closes #269 

Do not set the predictions in the dataset.

TODO:

- [x] check which tests are failing
- [x] check if we have to copy the predictions in the write folder when the dispatcher collect the results


glemaitre:
@lucyleeow @maikia apparently it works without further changes.
We only need to check where to copy the numpy arrays file.

PR #352, opened 2020-01-15
BUG disable event name update

maikia:
- It disables the option to change the event_name from the fronted
- if the event_name is already set wrong it displays the message to keep with the RAMP standards in the future.

1. **Before, if the event was wrongly named:** 
before_wrong

2. **Before, if the event was correctly named:**
before_correct

3. **After, if the event was wrongly named:**
after_wrong

4. **After, if the event was correctly named:**
after_correct

### DIFF
--- a/ramp-frontend/ramp_frontend/templates/update_event.html
+++ b/ramp-frontend/ramp_frontend/templates/update_event.html
@@ -38,25 +38,22 @@
          <div class="card-body">
            <form class="ui form" method="post" action="" name="update_event">
                {{ form.hidden_tag() }}
-                   {% if suffix_correct %}
-                        Event name will be {{ event.problem.name }} if you leave suffix empty, or {{ event.problem.name }}_&lt;suffix&gt;. The length should be less than 20 characters, only ascii, no space. It will be used to identify the event, primarily in the url. If you open several events for the same class or hackaton, please use the same suffix consistently.
-                        <div class="form-group">
-                            <label>event suffix</label>
-                            {{ form.suffix(placeholder="event_suffix") }}
-                            {% for error in form.suffix.errors %}
-                                <span style="color: red;">[{{ error }}]</span>
-                            {% endfor %}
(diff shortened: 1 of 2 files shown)


ramp-frontend/ramp_frontend/views/admin.py
```diff
@@ -201,13 +201,19 @@ def update_event(event_name):
                 .format(flask_login.current_user.name, event.name))
     admin = is_admin(db.session, event_name, flask_login.current_user.name)
     # We assume here that event name has the syntax <problem_name>_<suffix>
-    suffix = event.name[len(event.problem.name) + 1:]
+    # make sure that the event has the syntax <problem_name>_<suffix>
+    if event.name[:len(event.problem.name) + 1] == event.problem.name + '_':
```
glemaitre:
```suggestion
    is_suffix_correct = event.name.startswith(event.problem + '_'):
```

ramp-frontend/ramp_frontend/templates/update_event.html
```diff
@@ -37,15 +37,23 @@ <h4 class="modal-title" id="myModalLabel">{{ category }}</h4>
...
-                   <div class="form-group">
-                       <label>event suffix</label>
-                       {{ form.suffix(placeholder="event_suffix") }}
-                       {% for error in form.suffix.errors %}
-                           <span style="color: red;">[{{ error }}]</span>
-                       {% endfor %}
+                {% if not suffix_correct %}
+                    <div class="form-group"></div>
+                        <i class="fa fa-exclamation-triangle" aria-hidden="true"></i>
+                        Event name does not correspond to the RAMP standards. In the future, please make sure that it is preceeded by the 'problem_name_', eg: '{{ event.problem.name }}_{{ event.name }}'
```
glemaitre: Do we actually care to warn. Basically people will get a ValueError when deploying. I would just remove the field and that's it.

Issue #317, opened 2019-12-16
Write test for the view/sandbox in the frontend


glemaitre:
There is no test written for the frontend sandbox. It would be great to have this as an add-up.

PR #355, opened 2020-01-16
TST check file uploading in sandbox

maikia:
#### Reference Issue

 #317 
tests saving the submission file in sandbox

### DIFF
--- a/ramp-frontend/ramp_frontend/tests/test_ramp.py
+++ b/ramp-frontend/ramp_frontend/tests/test_ramp.py
@@ -413,7 +413,7 @@ def test_ask_for_event_mail(client_session):
 
 # add fixture to add and then remove files 
 # def submission_f_name()
-# file_dir, file_name, correct 
+# file_dir, file_name, part_of_code, correct 
 def test_sandbox_upload_file(client_session, makedrop_event): #, submission_f_name):
     client, session = client_session
 
@@ -426,6 +426,7 @@ def test_sandbox_upload_file(client_session, makedrop_event): #, submission_f_na
     from werkzeug.datastructures import ImmutableMultiDict
     from ramp_database.tools.user import get_user_interactions_by_name
     from ramp_database.model import SubmissionFile
+    from ramp_database.tools.submission import get_submission_by_name
     correct = True
 
     config = ramp_config_template()
@@ -451,8 +452,24 @@ def test_sandbox_upload_file(client_session, makedrop_event): #, submission_f_na
         assert rv.status_code == 302
         assert rv.location == 'http://localhost/events/iris_test_4event/sandbox'
 
-        user_interactions = get_user_interactions_by_name(session, 'test_user')
+        # check if the file was saved correctly
+        #user_interactions = get_user_interactions_by_name(session, 'test_user')
         if correct:
+            event = get_event(session, 'iris_test_4event')
+            sandbox_submission = get_submission_by_name(
+                                    session, 'iris_test_4event', 'test_user',
+            event.ramp_sandbox_name
+             )
+            submission_code = sandbox_submission.files[-1].get_code()
+
+            # check if the code of the submitted file is in the 'submission_code'
+            with open('data.txt', 'r') as file:
+                data = file.read()
+            import pdb; pdb.set_trace()
+            str(open(path_submission, 'rb')) in submission_code
+
+
+            import pdb; pdb.set_trace()
             assert user_interactions[user_interactions['interaction'] == 'upload']
             submission_file = \
                  (SubmissionFile.query.filter_by(submission=submission,
@@ -460,7 +477,7 @@ def test_sandbox_upload_file(client_session, makedrop_event): #, submission_f_na
                              .one_or_none())
 
             submission_file.get_code()
-        1/0
+            1/0
         else:
             assert not user_interactions[user_interactions['interaction'] == 'upload']
         import pdb; pdb.set_trace()
--- a/ramp-frontend/ramp_frontend/tests/test_ramp.py
+++ b/ramp-frontend/ramp_frontend/tests/test_ramp.py
@@ -413,46 +413,44 @@ def test_ask_for_event_mail(client_session):
 
 # add fixture to add and then remove files 
 # def submission_f_name()
+# file_dir, file_name, correct 
 def test_sandbox_upload_file(client_session, makedrop_event): #, submission_f_name):
     client, session = client_session
 
     #event = get_event(session, 'iris_test_4event')
-
     #session.commit()
     sign_up_team(session, 'iris_test_4event', 'test_user')
 
-    import pdb; pdb.set_trace()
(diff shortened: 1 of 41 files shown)


ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -411,6 +413,119 @@ def test_ask_for_event_mail(client_session):
                         in outbox[0].body)
 
 
+@pytest.mark.parametrize(
+    "submission_dir, file_name, correct",
```
glemaitre:
```suggestion
    "submission_dir, filename, correct",
```

glemaitre: I would write 2 separate tests. One for when we expect an error and one when we expect to work.

ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -411,6 +413,119 @@ def test_ask_for_event_mail(client_session):
...
+                             submission_dir, file_name, correct):
+    client, session = client_session
+    sign_up_team(session, 'iris_test_4event', 'test_user')
+
+    config = ramp_config_template()
+    ramp_config = generate_ramp_config(read_config(config))
+
+    # upload file in sandbox.html
+    path_submissions = os.path.join(
+                                    ramp_config['ramp_kit_dir'],
```
glemaitre:
```suggestion
        ramp_config['ramp_kit_dir'], submission_dir
```

ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -411,6 +413,119 @@ def test_ask_for_event_mail(client_session):
...
+                                    submission_dir
```
glemaitre:
```suggestion
```

ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -411,6 +413,119 @@ def test_ask_for_event_mail(client_session):
...
+                                   )
```
glemaitre:
```suggestion
    )
```

ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -411,6 +413,119 @@ def test_ask_for_event_mail(client_session):
...
+                             follow_redirects=False)
+
+            assert rv.status_code == 302
+            assert rv.location == 'http://localhost/events/' \
+                                  'iris_test_4event/sandbox'
+
+            # code of the saved file
+            with open(path_submission, 'r') as file:
+                submitted_data = file.read()
+        except FileNotFoundError:
```
glemaitre: What is the reason for the FileNotFoundError?

Basically I would expect that if the file is not found, we would like to check that we raise this error instead of using `except`.

ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -411,6 +413,119 @@ def test_ask_for_event_mail(client_session):
...
+                             + 'iris_test_4event/sandbox',
+                             headers={'Referer':
+                                      'http://localhost/events/'
+                                      + 'iris_test_4event/sandbox'},
+                             data={'file': (open(path_submission, 'rb'),
+                                            file_name)},
+                             follow_redirects=False)
+
+            assert rv.status_code == 302
+            assert rv.location == 'http://localhost/events/' \
```
glemaitre:
```suggestion
            assert (rv.location == 
                    'http://localhost/events/iris_test_4event/sandbox')
```

ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -411,6 +413,119 @@ def test_ask_for_event_mail(client_session):
...
+
+    with login_scope(client, 'test_user', 'test') as client:
+        rv = client.get('http://localhost/events/iris_test_4event/sandbox')
+        assert rv.status_code == 200
+
+        # choose file and check if it was uploaded correctly
+        path_submission = os.path.join(path_submissions, file_name)
+
+        try:
+            rv = client.post('http://localhost/events/'
```
glemaitre:
```suggestion
            rv = client.post(
                'http://localhost/events/iris_test_4event/sandbox',
                headers={
                    'Referer':
                    'http://localhost/events/iris_test_4event/sandbox'
                },
                data={
                    'file': (open(path_submission, 'rb'), filename)
                },
                follow_redirects=False
            )
```

ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -411,6 +413,119 @@ def test_ask_for_event_mail(client_session):
...
+
+            # code of the saved file
+            with open(path_submission, 'r') as file:
+                submitted_data = file.read()
+        except FileNotFoundError:
+            submitted_data = None
+
+        # code from the db
+        event = get_event(session, 'iris_test_4event')
+        sandbox_submission = get_submission_by_name(session,
```
glemaitre:
```suggestion
        sandbox_submission = get_submission_by_name(
            session, 'iris_test_4event', 'test_user', event.ramp_sandbox_name
        )
```

ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -411,6 +413,119 @@ def test_ask_for_event_mail(client_session):
...
+    client, session = client_session
+    sign_up_team(session, 'iris_test_4event', 'test_user')
+
+    example_code = 'example content'
+
+    with login_scope(client, 'test_user', 'test') as client:
+        rv = client.get('http://localhost/events/iris_test_4event/sandbox')
+        assert rv.status_code == 200
+
+        rv = client.post('http://localhost/events/'
```
glemaitre: If you could make the same change that I did above (this is the black python style).

agramfort:
what was the reason for running black here? it makes the diffs very long and harder to review. We agree to use black on the problem it should be on the CI as a check to avoid eternal diffs with cosmit changes. I personally prefer to agree on a set of warnings we ignore in setup.cfg and let flake8 tell us what is not ok. Then let's not make changes that are not raising issues with flake8

maikia:
> what was the reason for running black here? it makes the diffs very long and harder to review. We agree to use black on the problem it should be on the CI as a check to avoid eternal diffs with cosmit changes. I personally prefer to agree on a set of warnings we ignore in setup.cfg and let flake8 tell us what is not ok. Then let's not make changes that are not raising issues with flake8

ok. I reversed black from everywhere except those few cases pointed by @glemaitre 

What do you think of it now @glemaitre @agramfort ?

glemaitre:
@maikia Thanks for testing this spaghetti code. I might try to add up the test for the submission but at least we are covering something which was not the case before. This is good.

agramfort:
really nice. thx @maikia !

PR #362, opened 2020-01-18
ENH remove warnings from log

### LINKED ISSUE #209: Use new `ramp test` CLI in worker
We should use the new `ramp test --submission` instead of `ramp_test_submission` in the workshop.

### LINKED ISSUE #361: Remove warning from the log
In the dispatcher, I think that we should launch the worker without raising any warning.
They will only be available for us and are just noisy.

glemaitre:
closes #361 
closes #209
requires https://github.com/paris-saclay-cds/ramp-workflow/pull/208

Update the way `ramp_test_submission` is called using the new CLI in ramp-workflow

### DIFF
--- a/ramp-engine/ramp_engine/local.py
+++ b/ramp-engine/ramp_engine/local.py
@@ -130,7 +130,7 @@ class CondaEnvWorker(BaseWorker):
         environment given in the configuration. The submission is launched in
         a subprocess to free to not lock the Python main process.
         """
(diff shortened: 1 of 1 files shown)


glemaitre:
I tested locally and it should depend on ramp-workflow so we should be fine merging once 
https://github.com/paris-saclay-cds/ramp-workflow/pull/208 is merged.

ping @agramfort

PR #366, opened 2020-01-20
ENH add queue position for each submission

### LINKED ISSUE #277: display submission state in my_submissions view
Most students complain about lack of info about submission after submitting it. It would be rather easy to add a "state" column in the "New submissions (pending training)" table, so at least they get the detailed state if they refresh the page.

glemaitre:
closes #277 

Add information regarding the queue position of a submission


glemaitre:
But maybe we can wait since we need migration :)
At least we should stop the current running dispatcher.

rth:
Merging, thanks. We will have to do a DB migration for this release anyway.

PR #379, opened 2020-02-01
MNT add dispatcher parameters inside the config file

glemaitre:
We can move the parameter of the dispatcher inside the event config file since we are triggering a dispatcher by event. It will be much simpler to launch them then.

### DIFF
--- a/doc/create_ramp_event.rst
+++ b/doc/create_ramp_event.rst
@@ -40,6 +40,10 @@ This config file should look like::
     worker:
         worker_type: conda
         conda_env: ramp-iris
+    dispatcher:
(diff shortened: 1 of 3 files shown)


agramfort:
what I would expect as behavior is that we have default parameters for dispatcher for each event (eg 2 workers and 2 threads each) but that config can allow you to overwrite these values. is that we is done here?

glemaitre:
Yes apart that the default of worker that can be started started is the number of CPU and we don't limit the number of openmp threads. It was the previous default but we could change to a more sensible default (at least for the workers)

PR #382, opened 2020-02-02
ENH add daemon class to be able to start dispatchers for all open events

glemaitre:
This PR intends to add a daemon which will launch dispatcher inside subprocesses which can be killed all at once with a `KeyboardInterruption`.

### DIFF
--- a/ramp-engine/ramp_engine/cli.py
+++ b/ramp-engine/ramp_engine/cli.py
@@ -25,16 +25,8 @@ def main():
 @click.option("--event-config", show_default=True,
               help='Configuration file in YAML format containing the RAMP '
               'event information.')
-@click.option('--n-workers', default=-1, show_default=True,
-              help='Number of worker to start in parallel.')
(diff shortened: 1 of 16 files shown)


ramp-engine/ramp_engine/cli.py
```diff
@@ -18,6 +19,33 @@ def main():
...
 
 
+@main.command()
+@click.option("--config", default='config.yml', show_default=True,
+              help='Configuration file in YAML format containing the database '
+              'information.')
+@click.option("--events-dir", show_default=True,
+              help='Directory where the event config files are located.')
+@click.option('-v', '--verbose', count=True)
+def daemon(config, events_dir, verbose):
```
glemaitre: Note for review: This is a bit tricky to test because we can only kill it with a keyboard interruption

glemaitre:
@agramfort I think the PR is in good shape. I just want to check the tests coverage. I think that it should be almost all good apart of the CLI but we cannot test easily.

PR #396, opened 2020-02-10
ENH add fetcher for OSF data

glemaitre:
If we start to store data in OSF, we will need to fetch data from there.
We also need support for injecting private `token` since the data will be private.


glemaitre:
TODO:

- [ ] Write some test event. Not that it was live tested with: https://github.com/ramp-data/air_passengers/pull/1

glemaitre:
ping @agramfort

agramfort:
a test will tell me if it's good :)

glemaitre:
I merging after making a test for it.

PR #397, opened 2020-02-13
BUG allow user to download a submission

glemaitre:
Make the download button work


agramfort:
merge if happy @glemaitre !

PR #405, opened 2020-02-28
WIP Add navbar problem page

### LINKED ISSUE #400: add tab challenge leaderboard
#### Description

It would be of interest to have combined leaderboard from all the events run for one challenge.
This would be visible to all the users but the links to the actual solutions should be deactivated.
If one user takes part in more than one event only his/her best solution score should be displayed

lucyleeow:
closes #400 
Adds navbar to the top of problem page

### DIFF
--- a/ramp-frontend/ramp_frontend/static/css/themes/flat-blue.css
+++ b/ramp-frontend/ramp_frontend/static/css/themes/flat-blue.css
@@ -1121,7 +1121,7 @@
 }
 @media (min-width: 1200px) {
   .event-container {
-    width: 1100px;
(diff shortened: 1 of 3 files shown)


lucyleeow:
The navbar in the `event.html` and `problem.html` pages are the same length as the notebook iframe below it (and also the same width across the 2 pages).

Add navbar with 'description' and 'download kit' buttons to navbar:

image

Still to add leaderboard button linking to combined leaderboard for all events, with solutions deactivated unless it is the user's own solution.
I will work on this last part once @glemaitre is back.

PR #408, opened 2020-02-29
BUG solve issue by calling is_accessible_code with submission ID

### LINKED ISSUE #406: Issue with the "join event" button
It appears that under some version of chrome and firefox, the "join event" does not work.

glemaitre:
closes #406

### DIFF
--- a/doc/whats_new/v0.4.rst
+++ b/doc/whats_new/v0.4.rst
@@ -19,7 +19,7 @@ Changelog
   via the frontend.
   :pr:`397` by :user:`Guillaume Lemaitre <glemaitre>`.
 
-- Bug: fix a bug for which a user could not check is own code in
+- Bug: fix a bug for which a user could not check their own code in
   "My submission" menu.
   :pr:`408` by :user:`Guillaume Lemaitre <glemaitre>`.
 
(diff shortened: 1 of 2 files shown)


glemaitre:
ping @agramfort @lucyleeow @maikia We need to make a new 0.4 bug fix release if possible and deploy on the server. Just need to port this PR in 0.4.X branch then

glemaitre:
I might try to do that tonight.

glemaitre:
Nice failure, the test will fail every 4 years :)

doc/whats_new/v0.4.rst
```diff
@@ -19,5 +19,9 @@ Changelog
   via the frontend.
   :pr:`397` by :user:`Guillaume Lemaitre <glemaitre>`.
 
+- Bug: fix a bug for which a user could not check is own code in
```
lucyleeow:
```suggestion
- Bug: fix a bug for which a user could not check their own code in
```
'their' is gender neutral option (I know in French general is always masculine)

lucyleeow: Also I like that you drop the 'h' even when writing! :laughing:

glemaitre: :) no need for it

PR #410, opened 2020-03-04
BUG make a hard-copy when submitting the original kit

### LINKED ISSUE #409: Issue with external data symlink
When introducing the changes from @tomMoral (not to blame because it was not tested :)), I did not pay attention and we introduce a big regression.

When a user sign-up to an event we make a copy of the starting kit and submit it (for setting up the sandbox). If the external data were inside, we copied this file. Now with the symlink, we synlink to the starting kit and at every submission, we symlink the external data to this first submission data which symlink to the starting-kit external data.

Now when a user upload a file, it will be copied in place of the external data of the original kit where all the submissions are pointing to.

So the resolution is to make a hard-copy of the original external data when submitting the starting-kit. When uploading the data, we should replace the first submission and not the original data. We can check the diff on the `ramp.studio` server in production where I programmed this way in the past (including a parameter `hard_copy=True` for the original submission).

Be aware that this way one cannot download the external data from a previous submission because it will only be representative of the last submission uploaded.

glemaitre:
closes #409 

Make sure to copy external data when doing the first submission. All subsequent submissions will make symlink.

### DIFF
--- a/ramp-database/ramp_database/tools/submission.py
+++ b/ramp-database/ramp_database/tools/submission.py
@@ -39,7 +39,7 @@ logger = logging.getLogger('RAMP-DATABASE')
 # Add functions: add information to the database
 # TODO: move the queries in "_query"
 def add_submission(session, event_name, team_name, submission_name,
-                   submission_path):
+                   submission_path, hard_copy=False):
     """Create a submission in the database and returns an handle.
 
     Parameters
@@ -56,6 +56,9 @@ def add_submission(session, event_name, team_name, submission_name,
         The path of the files associated to the current submission. It will
         corresponds to the key `ramp_kit_subissions_dir` of the dictionary
         created with :func:`ramp_utils.generate_ramp_config`.
+    hard_copy : bool, default=False
+        Whether or not to trigger a hard copy or just a symlink of the external
+        data.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


glemaitre:
I tested locally and it will solve the current bug. However, we have some regression:

* When looking at a model, a user can get the external data. However, since we are always changing the original file, it means that we cannot anymore allow a user to inspect previous external data.

@agramfort @rth any thoughts on this issue

We should be able to add some quick regression test by checking if the external data are a symlink or not depending of the submission type.

agramfort:
I don't get it. Are you saying that new submission from a user will
overwrite the external data from his past submissions?

glemaitre:
Yes this the side effect of using symlink


                                                                                                                                      
 

 
     

agramfort:
but we should only use symlinks for FAN where there is a fixed hard coded
version of the external data.
I feel I am missing something

>

glemaitre:
This is not what the changes were doing or more precisely what is_editable was intended for




glemaitre:
To make it simple, `'data'` format type was declared to be a non-editable format and CSV is a from this `'data'` type.

rth:
Migration is indeed not easy, I'm not sure all of it can be automated. We should merge this in any case I think?

Personally to solve this issue, I am currently,
 - running RAMP with this PR, and a patch `add_submission(.. hard_copy=True)` by default just to be safe; for air passengers it's not too important anyway
 - had to replace the external_data.csv in the starting kit with the original file from Github (since that one was corresponded to a user submission)
 - will ask students who had issues with external data to re-submit.

glemaitre:
After discussing with @agramfort, it seems that the FAN challenge should not have external data but data which should be downloaded on the fly. Therefore, we can revert and always make copy. If we are facing a bigger issue in the future then we will need to touch the database most probably to have a viable solution. I will open a new PR.

rth:
OK, but maybe we can still merge this? Doesn't it partially solve the issue for other challenges with external data?

glemaitre:
#413 will solve the issue as well.

glemaitre:
it will just be a bit more costly space wise

PR #412, opened 2020-03-10
BUG sign up not working correctly

### LINKED ISSUE #404: Improve sign-up UX
The current user experience for signing up to RAMP could be improved. I just tried to create an account at https://dev.ramp.studio,

 - upon filing all the information in the registration form I was redirected to home page with no additional message about registration being successful or not.
 - trying to login produces: "Login Error: Wrong password"
 - trying to reset password produces: "The email address is not linked to any user. You can sign-up instead."  

I didn't recieve any messages from RAMP either. Some of those error messages should probably say that the account is pending manual validation. Or maybe the registration failed -- not sure.

User: rth
Email: indicated in my GH account (@rth)

maikia:
#### Reference Issue

Fixes #404
Now on the sign up the user is redirected to the index page without any message. The e-mail is send and the message pops up only if the user clicks sign up afterwards

This must be happening righ after the user signs up, before he is redirected (probably loading the page is now too fast and must be slown down as in the previous version)

### DIFF
--- a/doc/whats_new/v0.4.rst
+++ b/doc/whats_new/v0.4.rst
@@ -23,5 +23,9 @@ Changelog
   "My submission" menu.
   :pr:`408` by :user:`Guillaume Lemaitre <glemaitre>`.
 
+- Bug: fix a bug which was not showing modal box during sign-in due to lack
+  of redirection.
+  :pr:`412` by :user:`Maria Telenczuk <maikia>`.
+
 `ramp-utils`
 ............
--- a/ramp-frontend/ramp_frontend/tests/test_auth.py
+++ b/ramp-frontend/ramp_frontend/tests/test_auth.py
@@ -176,7 +176,7 @@ def test_sign_up(client_session):
     user_profile = {'user_name': 'xx', 'password': 'xx', 'firstname': 'xx',
                     'lastname': 'xx', 'email': 'xx'}
     rv = client.post('/sign_up', data=user_profile)
-    assert rv.status_code == 200
+    assert rv.status_code == 302
     user = get_user_by_name(session, 'xx')
     assert user.name == 'xx'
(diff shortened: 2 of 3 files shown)


maikia:
In the end the problem was that in the page to which the user was redirected to (ie index) does not handle the messages, so they were displayed only after a user went to the page which handles the messages. 
Instead of redirecting to index.html the user is redirected either to:
sign_up.html in case there was a problem with sigining up
login.html in case everything went fine
Both of those handle messages so it should now work fine

rth:
Thanks for investigating this @maikia !

rth:
I'm not too familiar with redirects in flask forms but the fix sounds reasonable to me.

rth:
Happy to merge it you can confirm that it's working as expected locally. Someone would also need to deploy this fix on dev.ramp.studio..

maikia:
Locally it works for the messages.

But there is one more thing. After signing up, even if you didn't see the message you should have gotten a confirmation e-mail. After confirming your e-mail you should have been able to use the account anyway.

Did you receive any e-mail?

rth:
> But there is one more thing. After signing up, even if you didn't see the message you should have gotten a confirmation e-mail. After confirming your e-mail you should have been able to use the account anyway.

> Did you receive any e-mail?

No, I didn't receive emails. And after I tried to sign-up the same thing happened (i.e. no message about user already existing). So I think my form was not properly sent/recorded in the database. If you have access to the server could you please check in the user table ? Credentials used are in the original issue. 

I'm not sure of fixing the redirect would address that.

maikia:
It's weird, but the e-mail would be send only when the message pops up (ie @rth  if you went afterwards to the page such as log in the e-mail would be send and the message would pop up and your request would be saved in the database.)

It should now work as expected

glemaitre:
Thanks @maikia I think this is good to be merged and back-ported in 0.4.X

PR #413, opened 2020-03-13
BUG always make copy of external data

### LINKED ISSUE #409: Issue with external data symlink

glemaitre:
closes #409 

Always make a copy instead of trying to symlink. We should go from the principle that the data are always quite small.

### DIFF
--- a/doc/whats_new/v0.4.rst
+++ b/doc/whats_new/v0.4.rst
@@ -9,6 +9,10 @@ Changelog
 `ramp-database`
 ...............
 
+- Bug: fix a bug which was always doing a symlink of the external data.
+Instead, we now always copy the file.
+:pr:`413` by :user:`Guillaume Lemaitre <glemaitre>`.
+
 `ramp-engine`
 .............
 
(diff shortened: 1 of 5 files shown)


glemaitre:
ping @agramfort @rth

glemaitre:
This PR needs to be back ported in 0.4.X for a bug fix release

agramfort:
Looks good. Can we still put this code on server as soon as the FAN challenge is closed?

glemaitre:
> Looks good. Can we still put this code on server as soon as the FAN challenge is closed?

Can I close FAN now?

glemaitre:
or do you still have some activities happening there?

PR #423, opened 2020-03-23
BUG amend default public opening time

### LINKED ISSUE #422: Submissions code is visible by default for event participants
It looks like submissions code in the Leaderboard is accessible by default for all participants: I.e. they can click on the submission name and get to the code.

Previously I though only an admin could do it. Not sure if it's related to an event being "competitive". Under "Update Event" it says,

> Check if the event is competitive. In competitive events users may dedicate the submission with which they would like to enter in the competition. It is not mandatory: if there is no dedicated submission, the best one according to the public leaderboard score will enter in the competition. 

In any case selecting or not that checkbox doesn't have an effect: the code is always visible.

That's probably not ideal, particularly for teaching events.

lucyleeow:
closes #422

### DIFF
--- a/ramp-database/ramp_database/model/event.py
+++ b/ramp-database/ramp_database/model/event.py
@@ -152,7 +152,7 @@ class Event(Model):
         DateTime, default=datetime.datetime(2000, 1, 1, 0, 0, 0))
     # before links to submissions in leaderboard are not alive
     public_opening_timestamp = Column(
-        DateTime, default=datetime.datetime(2000, 1, 1, 0, 0, 0))
+        DateTime, default=datetime.datetime(2100, 1, 1, 0, 0, 0))
     closing_timestamp = Column(
         DateTime, default=datetime.datetime(2100, 1, 1, 0, 0, 0))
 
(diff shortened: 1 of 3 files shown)


rth:
The `test_is_accessible_leaderboard` test seem to be failing..

lucyleeow:
Yes, I suspected there would be a failing - looking into it now!

lucyleeow:
I think the following test case in [`test_is_accessible_leaderboard`](https://github.com/paris-saclay-cds/ramp-board/blob/71bc6253a1cc06439700124f77e1532ee194c003/ramp-database/ramp_database/tools/tests/test_frontend.py#L180) is wrong, now that we have changed the default open time:
```
    # simulate a user which authenticated and author of the submission to a
    # public event
    user.is_authenticated = True
    assert is_accessible_leaderboard(session_toy_db, event_name, user.name)
```
Just for reference, this case has the following settings and should return False:
* user.is_authenticated = True
* user.is_active = True (actually always true and redundant)
* is_admin = False
* is_user_signed_up = True
* is_public_open (checks if current time is between public open and closing i.e. collaboration phase) = False (previously True)

Note `is_accessible_leaderboard` determines if the submission code is accessible on the leaderboard ('the submission name is clickable'), which should be no in the above circumstance.

PR #429, opened 2020-06-17
WIP DOC Update aws worker

maikia:
explanation on how to set up the AWS worker updated

### DIFF
--- a/doc/workers.rst
+++ b/doc/workers.rst
@@ -69,6 +69,10 @@ Running submissions on Amazon Web Services (AWS)
 The AWS worker will train and evaluate the submissions on an AWS instance,
 which can be useful if the server itself has not much resources or if you need
 specific resources (e.g. GPU's).
+to create an instance:
+when you are saving the key pair put it in the secure location (eg your .ssh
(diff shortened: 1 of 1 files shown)


doc/workers.rst
```diff
@@ -105,45 +142,87 @@ A very short how-to for creating such an AMI manually:
...
           worker_type: aws
           access_key_id: <aws_access_key_id for boto3 Session>
           secret_access_key: <aws_secret_access_key for boto3 Session>
           region_name: us-west-2 # oregon
           ami_image_name: <name of the AMI set up for this event>
-          ami_user_name : ubuntu
-          instance_type : t2.micro
+          ami_user_name: ubuntu
+          instance_type: t2.micro
           key_name: <user that can ssh connect with the created AWS instance>
```
lucyleeow, line 183: Again I think this is the name of the key not the name of the user ?

lucyleeow: I'm not 100% sure that the key_name is the name of the ssh key, because my text ultimately failed. Did you double check this on your machine...?

PR #435, opened 2020-07-02
ENH Add dispatcher log

### LINKED ISSUE #434: ENH Save dispatcher logs
I don't think dispatcher logs are being saved atm (just submission logs) and this would be useful, esp with AWS workers.

lucyleeow:
#### Reference Issue
Closes #434

#### Any other comments?

ping @agramfort @maikia

### DIFF
--- a/db_engine.yml
+++ b/db_engine.yml
@@ -1,2 +1,2 @@
-db_owner: lucy # to find out your db_owner type psql -l in the terminal
+db_owner: postgres # to find out your db_owner type psql -l in the terminal
                    # and check who is the owner of your postgres database 
--- a/db_engine.yml
+++ b/db_engine.yml
@@ -1,2 +1,2 @@
-db_owner: postgres # to find out your db_owner type psql -l in the terminal
+db_owner: lucy # to find out your db_owner type psql -l in the terminal
                    # and check who is the owner of your postgres database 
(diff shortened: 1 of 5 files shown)


db_engine.yml
```diff
@@ -1,2 +1,2 @@
-db_owner: postgres # to find out your db_owner type psql -l in the terminal
+db_owner: lucy # to find out your db_owner type psql -l in the terminal
```
maikia, line 1:
```suggestion
db_owner: postgres  # to find out your db_owner type psql -l in the terminal
```

ramp-engine/ramp_engine/base.py
```diff
@@ -3,6 +3,17 @@
 
 logger = logging.getLogger('RAMP-WORKER')
 
+log_file = "worker.log"
```
glemaitre, line 6: I have 2 questions here:

* we are always logging in the same file whatever and whenever a worker is started?
* If the log start to be too long it will be a problem. Thus the current config take care about splitting the file and renaming it something like `worker.log.1`, ...?

glemaitre: Basically something like that: https://stackoverflow.com/questions/39442558/logging-rewrite-if-size-gets-too-big

lucyleeow: > we are always logging in the same file whatever and whenever a worker is started?

I am not sure about the worker. It looks like it would be quite difficult to to deal with this: https://stackoverflow.com/questions/15096090/python-separate-processes-logging-to-same-file . It depends on how useful the worker logs are, for now it seems the dispatcher log is of most use.

PR #437, opened 2020-07-09
Use rotating log files

lucyleeow:
#### Reference Issue
Use a `RotatingFileHandler` to write to new log file once size limit met, as suggested by @glemaitre in #435.

#### Any other comments?

cc @maikia @agramfort

### DIFF
--- a/ramp-engine/ramp_engine/aws/worker.py
+++ b/ramp-engine/ramp_engine/aws/worker.py
@@ -1,4 +1,5 @@
 import logging
+from logging.handlers import RotatingFileHandler
 import os
 
 from ..base import BaseWorker, _get_traceback
@@ -13,9 +14,7 @@ logger = logging.getLogger('RAMP-AWS')
 
 log_file = os.path.join(log_dir, 'aws_worker.log')
 formatter = logging.Formatter('%(asctime)s %(name)s %(levelname)s %(message)s')  # noqa
-fileHandler = logging.handlers.RotatingFileHandler(
-    log_file, maxBytes=1000, backupCount=2
-)
+fileHandler = .RotatingFileHandler(log_file, maxBytes=1000, backupCount=2)
 fileHandler.setFormatter(formatter)
 streamHandler = logging.StreamHandler()
 streamHandler.setFormatter(formatter)
--- a/ramp-engine/ramp_engine/aws/worker.py
+++ b/ramp-engine/ramp_engine/aws/worker.py
@@ -1,14 +1,23 @@
 import logging
+from logging.handlers import RotatingFileHandler
+
+import os
 
 from ..base import BaseWorker, _get_traceback
 from . import api as aws
 
 
+log_dir = "logs"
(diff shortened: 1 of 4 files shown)


ramp-engine/ramp_engine/dispatcher.py
```diff
@@ -30,11 +31,15 @@
...
+if not os.path.exists(log_dir):
+    os.makedirs(log_dir)
+
 logger = logging.getLogger('RAMP-DISPATCHER')
 
-log_file = 'dispatcher.log'
+log_file = os.path.join(log_dir, 'dispatcher.log')
 formatter = logging.Formatter('%(asctime)s %(name)s %(levelname)s %(message)s')  # noqa
-fileHandler = logging.FileHandler(log_file, mode='a')
+fileHandler = RotatingFileHandler(log_file, maxBytes=1000, backupCount=2)
```
agramfort, line 42:
```suggestion
file_handler = RotatingFileHandler(log_file, maxBytes=1000, backupCount=2)
```

ramp-engine/ramp_engine/dispatcher.py
```diff
@@ -30,11 +31,15 @@
...
 fileHandler.setFormatter(formatter)
 streamHandler = logging.StreamHandler()
```
agramfort, line 44:
```suggestion
stream_handler = logging.StreamHandler()
```

agramfort (commented):
besides my pep8 nitpicks LGTM

ramp-engine/ramp_engine/aws/worker.py
```diff
@@ -14,13 +14,13 @@
 
 log_file = os.path.join(log_dir, 'aws_worker.log')
 formatter = logging.Formatter('%(asctime)s %(name)s %(levelname)s %(message)s')  # noqa
-fileHandler = RotatingFileHandler(log_file, maxBytes=1000, backupCount=2)
-fileHandler.setFormatter(formatter)
+file_handler = RotatingFileHandler(log_file, maxBytes=1000, backupCount=2)
+file_handler.setFormatter(formatter)
 streamHandler = logging.StreamHandler()
 streamHandler.setFormatter(formatter)
```
agramfort, line 20: streamHandler -> stream_handler

agramfort (approved):
@glemaitre green button is all yours

ramp-engine/ramp_engine/base.py
```diff
@@ -1,18 +1,24 @@
 import logging
+from logging.handlers import RotatingFileHandler
+import os
 from abc import ABCMeta, abstractmethod
 
+log_dir = "logs"
```
rth, line 6: Can't we make this configurable somehow? 
The issue is that this log folder will depend on the location from which we run a script as far as I understand (an also happens at import). It might be better to have a function that is called as necessary that takes as input a config object and returns this logger. Not sure how easy that would be though.

PR #401, opened 2020-02-20
ENH Delete predictions

maikia:
#### Reference Issue

closes #399 

The user might be interested to remove the predictions from events (predictions may take a lot of disk space) while leaving the rest of the information unchanged.

This code enables the user to do so from the terminal

PR #449, opened 2020-10-20
ADV Add optional contributivity calculations

rth:
@agramfort @lucyleeow @glemaitre @maikia Generally feel free to skip/not review all PRs with the `ADV` tag in the title. Those are made the `advanced` branch, not `master`. Though if you like something we can certainly backport it to master as well.
 
Extracts part of https://github.com/paris-saclay-cds/ramp-board/pull/276

>  - Optional credit tracking, set by the "TRACK_CREDITS" config variable. Credit tracking is needed if we want to compute historical contributivities.
> - CLI command ramp-database compute-contributivity --event --ramp-kit-dir --ramp-data-dir. It computes contributivity and historical contributivity scores (displayed in two leaderboard columns) and the combined score (displayed at the top of the leaderboard pages). It is not run automatically as before, admins need to update it time to time. Running time is quadratic in number of submissions, and can be quite long of the data sets are big (more precisely: if the combination function is slow).

TODO
 - [x] fix path issue due to folder structure inconsistency between ramp-board and ramp-workflow


kegl:
Symlinking is a great patch but it makes it even more weird that we copy the predictions to somewhere else just to symlink them back. If anything, I'd symlink the other way around (instead of copying). We can merge this now and open an issue during/after the optical challenge. I'll just ping @glemaitre now, do you have a quick answer from your memory to why we copy the predictions from `submissions/submissions_<id>/training_output/` to `predictions/sumbmission_<id>/`?

rth:
>  now, do you have a quick answer from your memory to why we copy the predictions from submissions/submissions_<id>/training_output/ to predictions/sumbmission_<id>/?

Guillaume mentioned that it was likely initially done be able to easily remove all prediction files (to save space) while keeping the submissions. Though if we merge https://github.com/paris-saclay-cds/ramp-board/pull/401 that should be possible irrespective of the folder structure.

> Symlinking is a great patch but it makes it even more weird that we copy the predictions to somewhere else just to symlink them back. If anything, I'd symlink the other way around (instead of copying). We can merge this now and open an issue during/after the optical challenge.

Yes, you are right, I'm not so keen on this symlinking in retrospective. For the local conda worker we could indeed just skip the [removal of the `training_output` dir](https://github.com/paris-saclay-cds/ramp-board/blob/f9698ebb73e4f53e22988c3acead0a855740d586/ramp-engine/ramp_engine/local.py#L102) and possibly symlink  to [predictions dir](https://github.com/paris-saclay-cds/ramp-board/blob/f9698ebb73e4f53e22988c3acead0a855740d586/ramp-engine/ramp_engine/local.py#L196). Though then it would need to also be done for other worker types.

For context the issue is that a separate prediction dir in ramp_board is not compatible with ramp_workflow assumptions of the folder structure, which assumes that the training output is located in `<submission_dir>/<submission_id>/training_output`. However this folder is currently removed by the worker after training..

agramfort:
do we have a way to migrate an existing server?

>

rth:
This branch should be backward compatible if we wanted to add it to master (i.e. it only adds a symlink) in each submission dir. If we wanted to apply some of the other choices mentioned in my previous comment, then yes we would need to write some migration script for the existing servers as well.

agramfort:
so it means we don't touch the ramp.studio server?

>

rth:
> so it means we don't touch the ramp.studio server?

This is a PR to the `advanced` branch, while ramp.studio is on the `master` branch (or more specifically using releases). So this does not affect the ramp.studio server for now. Whether you would be interested in backporting this or other features recently merged to `advanced` is up to you (if so I can make the PR). Of course if we don't sync these two branches somewhat, they risk to diverge with time.

PR #450, opened 2020-10-22
ADV/STY Round train/test times to ms in the leaderboard

rth:
Minor styling fix from https://github.com/paris-saclay-cds/ramp-board/pull/276/

PR to `advanced` but can be also backported to `master`

Rounds the timing shown in the leaderboard to ms, currently they are in rounded to µs (which is not relevant and probably  larger than the measurement error)


kegl:
Hm, I would round it to sec actually.

rth:
> Hm, I would round it to sec actually.

Yes, but prediction time are also lower than 1s in which case it would often show 0.  Also for small datasets (a few 100 samples) even training time can be quite fast with simple models..

kegl:
OK, let's keep it ms and merge, thanks a lot @rth !

Issue #444, opened 2020-10-08
Add a remote worker


rth:
Currently there is a local `CondaEnvWorker` and a remote `AWSWorker`. I would be usful to have a generic `RemoteWorker` that's not tied to EC2 API. The idea would be to make submissions to a remote server that's already running.

This could likely be done either,
 1. using dask distributed with 1 remote dask worker
 2. or possibly something like [celery + rabbitmq for communication](https://avilpage.com/2014/11/scaling-celery-sending-tasks-to-remote.html) (though this would likely be more difficult to setup than the first option).

agramfort:
+1 for trying with dask

>

kegl:
+1 also for dask. We had an early celery-based attempt that we ended up scraping. it was hard to manage when something went wrong.

glemaitre:
I think this is a great idea. I think that we wanted to try something with `OpenStack` as well at some point.
I think this is only by implementing a dask (or other) worker that we can see how much abstraction we can have in `RemoteWorker`.
`BaseWorker` already provides the protocol but I don't know how much more the `RemoteWorker` can do.

@rth do you have some insights?

rth:
So there is an initial implementation with a dask.distributed worker in https://github.com/paris-saclay-cds/ramp-board/pull/452. It passes all worker and dispatcher tests with a local dask cluster and most tests with a remote dask cluster (assuming the file paths for ramp kit, data etc are the same on the remote server). A few issues still  need to be ironed out, but it gives a general idea. The code structure is very similar to  a local conda worker.

PR #452, opened 2020-10-26
ADV Add DaskWorker

rth:
This is a WIP attempt to implement a remote worker with dask.distributed.

Closes https://github.com/paris-saclay-cds/ramp-board/issues/444

PR to the `advanced` branch

## Usage

To use this worker, one would have to,
 1. setup a ramp event on the local machine, and deploy the event
 2. copy `ramp-kits` and `ramp-data` folder to the remote machine, install ramp-board and the appropriate conda env there. The structure of the `ramp_deployment` folder the remote machine should be the same as on the local machjne.
 3. Start a dask scheduler on the local machine, and one or several workers on the remote machine (cf https://docs.dask.org/en/latest/setup/cli.html#command-line) or use [dask-ssh](https://docs.dask.org/en/latest/setup/ssh.html) tool to do both.
 4. In the event config, specify
     ```json
     worker:
         worker_type: remote
         conda_env: ramp-iris
         dask_scheduler: "tcp://127.0.0.1:8786"   # or another URL to the dask scheduler
     ````
  5. Start the dispatcher for the event.

Some of the last 3 steps might be aggregated/automated.

## Implementation

The general idea is that we do exactly the same steps as in the local conda folder, but
 - we wrap calls into a `dask.distributed.Client.submit` to execute them on the remote machine
 - for data copies, getting logs etc we need to explicitly serialize it between remote/local using dask.

The nice thing is that we can run a local dask cluster, and easily run unit tests  on it. Currently I'm running all the  CondaEnvWorker unit tests and also dispatcher tests, both with this local worker and remote worker (there are still some failures in the dispatcher that should be fixed). All data serialization is not yet done via dask either.

There are some code changes in the `ramp_engine/local.py` as well, since I factorized functions that could be re-used in the remote worker in `ramp_engine/conda.py`

## TODO
 - [x] fix failing dispatcher tests with RemoteWorker
 - [x] adapt all file manipulation calls to use dask
 - [x] ~improve path adaption between the local and remote paths. E.g. in the event config we need to decide whether remote path or local paths are provided for each of the entries.~ to keep it simple, data, submission and kit paths should be the same between remote and local machine.
 - [x] test with an actual dask worker on a remote machine -> generally it work, a few tests still need fixing. 
 - [x] add documentation section


glemaitre:
We will be interested in master as well ;)

rth:
Yes, I know, there is nothing specific to the advanced branch here. I can look into backporting to master once it's merged here )

glemaitre:
> copy ramp-kits and ramp-data folder to the remote machine, install ramp-board and the appropriate conda env there. The structure of the ramp_deployment folder the remote machine should be the same as on the local machjne.

In the AWS worker, it is part of the worker to upload on the remote machine the submission and then collect the results.
Here, it seems that you want to outsource this part to the user. Would it make sense to have the same pattern?

glemaitre:
Another thing would be that it could be the job of the dispatcher to start a sask scheduler if requested, isn't it?

ramp-engine/ramp_engine/conda.py
```diff
@@ -0,0 +1,68 @@
+# All the following functions may be used with dask, so
+# they should include imports inside functions.
+from typing import Dict
+
+
+def _conda_info_envs() -> Dict:
```
glemaitre, line 6: I am not using `virtualenv` but do you think that it would be possible to have a higher-level API that could unify either using conda or virtualenv (of course not in this PR and anyway all these stuff are private).

rth:
> In the AWS worker, it is part of the worker to upload on the remote machine the submission and then collect the results.

The collection of the result is done in https://github.com/paris-saclay-cds/ramp-board/pull/452/files#diff-4f94babe188ee4f21f182143a33262de7bcd8ab89ba9520e9b67e1ca3f3f4a5bR261 but you are right that I still to add equivalent code for uploading submissions in https://github.com/paris-saclay-cds/ramp-board/pull/452/files#diff-4f94babe188ee4f21f182143a33262de7bcd8ab89ba9520e9b67e1ca3f3f4a5bR200 (there is a TODO note).

> Another thing would be that it could be the job of the dispatcher to start a dask scheduler if requested, isn't it?

Well it could (cf https://docs.dask.org/en/latest/setup/python-advanced.html#scheduler). Though that code is blocking, and would need to be run in an additional subprocess, and I'm not sure what would happen to stdout log handling. Then one need to start workers on the remote manually in any case, unless one uses [dask SSH](https://docs.dask.org/en/latest/setup/ssh.html). That's something we can investigate once we are sure that this fully works as expected, but initially starting them both manually is probably safer.

rth:
> In the AWS worker, it is part of the worker to upload on the remote machine the submission and then collect the results.

In the future, we could even probably copy the ramp-data and ramp-kit folder from the local machine, it's not present on the remote. But we probably don't want to do it for each submission (only once) for performance reasons, and I guess copying manually initially shouldn't be too problematic.

glemaitre:
> we could even probably copy the ramp-data and ramp-kit folder from the local machine

This is true that in AWS, we get this done by deploying the image.

OK so we will have to make a choice here at some point:

* get a common base class with high-level API shared between AWS and the current remote worker;
* or having dissociated class which will be more effective in terms of performance.

Anyway, this is not that important for the moment. Having something working is better and then we can see internal refactoring after seeing the limitation in practice.

ramp-engine/ramp_engine/conda.py
rth: yeah, there is nothing too conda specific in workers. We just need to compute the path to the python executable inside the virtualenv. The only issue is that virtualenv doesn't have a CLI to list all available envs. So we would probably need to use some additional tool on top, maybe [poetry](https://python-poetry.org/docs/managing-environments/#listing-the-environments-associated-with-the-project) or [virtualenvwrapper](https://virtualenvwrapper.readthedocs.io/en/latest/command_ref.html?highlight=lsvirtualenv#lsvirtualenv) (unless we are OK passing the path to the virtualenv in the config).

glemaitre:
> Well it could (cf https://docs.dask.org/en/latest/setup/python-advanced.html#scheduler). Though that code is blocking, and would need to be run in an additional subprocess, and I'm not sure what would happen to stdout log handling. Then one need to start workers on the remote manually in any case, unless one uses dask SSH. That's something we can investigate once we are sure that this fully works as expected, but initially starting them both manually is probably safer.

I had a nice asynchronous non-blocking implementation for the dispatcher: https://gist.github.com/glemaitre/b0f0dddcadd9ebe7df39b657c777ac50 which was waiting on the shelves :)

> Then one need to start workers on the remote manually in any case, unless one uses dask SSH

OK so there is a little benefit there.

rth:
> I had a nice asynchronous non-blocking implementation for the dispatcher

Nice! Wouldn't it require for [workers to also be async](https://gist.github.com/glemaitre/b0f0dddcadd9ebe7df39b657c777ac50#file-asyncio-py-L44) though?

kegl:
The way we used AWS was to set up an image that contained everything except for the submission. Setting it up was semi-manual. We can have something formalized into some customizable script, what to do to set up the backend.

ramp-engine/ramp_engine/remote.py
```diff
@@ -0,0 +1,302 @@
...
+            pure=False,
+        )
+
+        super().launch_submission()
+        self._start_date = datetime.utcnow()
+
+    def collect_results(self):
+        """Collect the results after that the submission is completed.
+
+        Be aware that calling ``collect_results()`` before that the submission
```
albertcthomas, line 249: we could maybe add a check for this? if the worker.status is `running` we don't collect the results. If yes this can be refactored later maybe

ramp-engine/ramp_engine/tests/test_conda_worker.py
```diff
@@ -103,80 +115,96 @@ def test_conda_worker(submission, get_conda_worker):
...
+        del worker.config['kit_dir']
+
+        err_msg = "The worker required the parameter 'kit_dir'"
+        with pytest.raises(ValueError, match=err_msg):
+            worker.setup()
+
 
+@pytest.mark.parametrize("Worker", ALL_WORKERS)
+def test_conda_worker_error_unknown_env(Worker, dask_scheduler):
+    with get_conda_worker(
```
albertcthomas, line 150: maybe add a conda_env = 'xxx' variable used in the context manager declaration and in the msg_err

ramp-engine/ramp_engine/tests/test_remote_worker.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def test_dask_workers_single_machine():
+    workers_nthreads = {
+        'tcp://127.0.0.1:38901': 4,
+        'tcp://127.0.0.1:43229': 4,
+        'tcp://127.0.0.1:46663': 4
+    }
+
+    assert _check_dask_workers_single_machine(workers_nthreads.keys()) is True
```
albertcthomas, line 16: I don't think you need the `is True`

albertcthomas (commented):
I did not look at the dispatcher tests.

ramp-engine/ramp_engine/tests/test_remote_worker.py
rth: without it, any not None or not False object would pass the assert. Here I wanted to check that the output is explicitly `True`

ramp-engine/ramp_engine/remote.py
rth: Yes, that's what happens in the dispatcher, which handles when to call these methods depending on the worker status. However here this comment is still useful just to indicate the the call to `collect_results` is blocking.

ramp-engine/ramp_engine/remote.py
```diff
@@ -0,0 +1,302 @@
...
+def _remove_link_or_dir(path: str):
+    """Remove a path that could be either a folder or a symlink to a folder"""
+    if os.path.islink(path):
+        # path is a symbolic link
+        os.unlink(path)
+    else:
+        shutil.rmtree(path, ignore_errors=True)
+
+
+class RemoteWorker(BaseWorker):
```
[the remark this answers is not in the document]
rth: Good point. Renamed to `DaskWorker`, as it's not necessarily remote: it could be used with a local Dask cluster (as it's done in the tests).

ramp-engine/ramp_engine/tests/test_remote_worker.py
albertcthomas: right

rth:
Thanks a lot for the review @albertcthomas I addressed your comments.

> The way we used AWS was to set up an image that contained everything except for the submission. Setting it up was semi-manual. We can have something formalized into some customizable script, what to do to set up the backend.

Definitely, I agree @kegl we can certainly make this more user friendly. Both by using a pre-build Docker image for the worker and by automating more things (e.g. deployment of ramp-data, ramp-kit on the remote server. Or at least making sure they are consistent with the local server). As a first approach this works though, and I think, once I add a bit more docs, we can merge it into `advanced` since that's the version that's currently deployed for the event (and I would rather make PRs to advanced rather than this branch for other fixes).

.gitignore
```diff
@@ -25,4 +25,9 @@ test_submissions/submission*
 *.egg-info
 databoard.db
 db_engine.yml
+<<<<<<< HEAD
+prof
+dask-worker-space/
+=======
 coverage.xml
+>>>>>>> upstream/master
```
agramfort, line 33: Nice ;)

rth: Thanks! Not sure how it got there, didn't see any conflicts..

PR #453, opened 2020-10-27
[WIP] ADV Getting rid of predictions in the DB 

kegl:


rth:
LGTM, thanks @kegl . I checked this works on the server (even without a DB migration visibly). Good to remove unused code.

kegl:
Yes, it was already removed by @glemaitre I think.

glemaitre:
Yep, I think that I commented out the part that where storing the predictions but letting the code as is just in case that it was not a wise choice. I think that we did not get into trouble, so clean-up was a great thing.

rth:
Great!

PR #455, opened 2020-10-30
ENH Enable use of AWS spot instances

lucyleeow:
Use [`request_spot_instances`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.request_spot_instances).

#### Any other comments?
~Problem with `request_spot_instances` is that it returns a dictionary. Currently we use [`create_instances`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.ServiceResource.create_instances) which returns a `list(ec2.Instance)`, where [`ec2.instance`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Instance) is a class.~
Hopefully `list(resource.Instance(instance_id))` is the same type of object as returned by [`create_instances`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.ServiceResource.create_instances) which returns a `list(ec2.Instance)`.

All the current code is written for `ec2.instance`

cc @agramfort @glemaitre

### DIFF
--- a/ramp-engine/ramp_engine/base.py
+++ b/ramp-engine/ramp_engine/base.py
@@ -62,6 +62,11 @@ class BaseWorker(metaclass=ABCMeta):
         self.status = 'killed'
 
     @abstractmethod
+    def _is_submission_terminated(self):
+        """Check if submission has been terminated."""
+        pass
+
+    @abstractmethod
     def _is_submission_finished(self):
         """Indicate the status of submission"""
         pass
@@ -71,7 +76,9 @@ class BaseWorker(metaclass=ABCMeta):
         status = self._status
         if status == 'running':
             self._status_running_check_time = datetime.utcnow()
-            if self._is_submission_finished():
+            if self._is_submission_terminated():
+                self._status = 'reset'
+            elif self._is_submission_finished():
                 self._status = 'finished'
         return self._status
 
--- a/ramp-engine/ramp_engine/base.py
+++ b/ramp-engine/ramp_engine/base.py
@@ -31,12 +31,13 @@ class BaseWorker(metaclass=ABCMeta):
     status : str
         The status of the worker. It should be one of the following state:
 
-            * 'initialized': the worker has been instanciated.
+            * 'initialized': the worker has been instantiated.
             * 'setup': the worker has been set up.
-            * 'error': setup failed / training couldn't be started
+            * 'error': setup failed / training couldn't be started.
             * 'running': the worker is training the submission.
             * 'finished': the worker finished to train the submission.
             * 'collected': the results of the training have been collected.
+            * 'retry': the worker has been interrupted (and will be retried).
             * 'killed'
     """
     def __init__(self, config, submission):
@@ -62,8 +63,8 @@ class BaseWorker(metaclass=ABCMeta):
         self.status = 'killed'
 
     @abstractmethod
-    def _is_submission_terminated(self):
-        """Check if submission has been terminated."""
+    def _is_submission_interrupted(self):
+        """Check if submission has been interrupted."""
         pass
 
     @abstractmethod
@@ -76,8 +77,8 @@ class BaseWorker(metaclass=ABCMeta):
         status = self._status
         if status == 'running':
             self._status_running_check_time = datetime.utcnow()
-            if self._is_submission_terminated():
-                self._status = 'reset'
+            if self._is_submission_interrupted():
+                self._status = 'retry'
             elif self._is_submission_finished():
                 self._status = 'finished'
         return self._status
--- a/ramp-engine/ramp_engine/dispatcher.py
+++ b/ramp-engine/ramp_engine/dispatcher.py
@@ -86,9 +86,6 @@ class Dispatcher:
            submissions, as the check for collection will be done through SSH.
            Thus, if the time between checks is too small, the repetitive
            SSH requests may be potentially blocked by the cloud provider.
-    use_spot_instance : bool, default=False
-        Whether or not to use AWS spot instance. Only relevant when using AWS
-        worker.
     """
     def __init__(self, config, event_config, worker=None, n_workers=1,
                  n_threads=None, hunger_policy=None,
--- a/ramp-engine/ramp_engine/dispatcher.py
+++ b/ramp-engine/ramp_engine/dispatcher.py
@@ -198,14 +198,20 @@ class Dispatcher:
                 self._processing_worker_queue.put_nowait(
                     (worker, (submission_id, submission_name)))
                 time.sleep(0)
+            elif worker.status == 'reset':
+                self._reset_single_submission(
+                    session, submission_id
+                )
+                logging.info(f'Submission: {submission_id} has been reset due '
+                             'to spot instance being terminated by AWS.')
+                worker.teardown()
             else:
                 logger.info(f'Collecting results from worker {worker}')
                 returncode, stderr = worker.collect_results()
                 if returncode:
                     if returncode == 124:
                         logger.info(
-                            'Worker {} killed due to timeout.'
-                            .format(worker)
+                            f'Worker {worker} killed due to timeout.'
                         )
                     else:
                         logger.info(
@@ -252,6 +258,10 @@ class Dispatcher:
             update_leaderboards(session, self._ramp_config['event_name'])
             update_all_user_leaderboards(session,
                                          self._ramp_config['event_name'])
+
+    @staticmethod
+    def _reset_single_submission(session, submission_id):
+        set_submission_state(session, submission_id, 'new')
 
     @staticmethod
     def _reset_submission_after_failure(session, even_name):
--- a/ramp-engine/ramp_engine/dispatcher.py
+++ b/ramp-engine/ramp_engine/dispatcher.py
@@ -198,12 +198,11 @@ class Dispatcher:
                 self._processing_worker_queue.put_nowait(
                     (worker, (submission_id, submission_name)))
                 time.sleep(0)
-            elif worker.status == 'reset':
-                self._reset_single_submission(
-                    session, submission_id
-                )
-                logging.info(f'Submission: {submission_id} has been reset due '
-                             'to spot instance being terminated by AWS.')
+            elif worker.status == 'retry':
+                set_submission_state(session, submission_id, 'new')
+                logging.info(f'Submission: {submission_id} has been '
+                             'interrupted. It will be added to queue again '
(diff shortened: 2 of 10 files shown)


lucyleeow:
@agramfort @glemaitre this is ready for review if either of you want to look at it. After this is merged I still need to make a patch release and update ramp-board on the server. Should probably be done before the course on Friday?

I've tried this on my local RAMP server a few times, sending submissions to AWS and it works fine.

I was not able to test:
* when the spot instance is not fulfilled (e.g., due to lack of capacity etc)
* when a spot instance is terminated (manually terminating an instance does not occur in the same way as AWS terminating a spot instance)

FYI @rth I have added a `abstractmethod` `_is_submission_terminated` to the `BaseWorker`, which returns whether or not a spot instance as been terminated for AWS worker, and for CondaWorker always returns False.

ramp-engine/ramp_engine/base.py
```diff
@@ -61,6 +61,11 @@ def teardown(self):
         """Clean up (i.e., removing path, etc.) before killing the worker."""
         self.status = 'killed'
 
+    @abstractmethod
+    def _is_submission_terminated(self):
```
rth, line 65: Maybe we could call it, `_is_worker_active` or something along those lines? Because the current name makes me think that the submission was cancelled. While it was the whole machine that was terminated/cancelled as far as I understand. Or at least some other way to differentiate between this and the `worker._status`. In the context of a `DaskWorker` https://github.com/paris-saclay-cds/ramp-board/pull/452 I guess that would be something like no workers found.

ramp-engine/ramp_engine/base.py
```diff
@@ -71,7 +76,9 @@ def status(self):
         status = self._status
         if status == 'running':
             self._status_running_check_time = datetime.utcnow()
-            if self._is_submission_finished():
+            if self._is_submission_terminated():
+                self._status = 'reset'
```
rth, line 80: Maybe ?

```suggestion
                self._status = 'retry'
```

ramp-engine/ramp_engine/dispatcher.py
```diff
@@ -198,14 +198,20 @@ def collect_result(self, session):
                 self._processing_worker_queue.put_nowait(
                     (worker, (submission_id, submission_name)))
                 time.sleep(0)
+            elif worker.status == 'reset':
+                self._reset_single_submission(
+                    session, submission_id
+                )
+                logging.info(f'Submission: {submission_id} has been reset due '
+                             'to spot instance being terminated by AWS.')
```
rth, line 206: Maybe only log this when
```
isinstance(worker, AWSWorker)
```
and otherwise print a more generic error about the submission training being interrupted and that it will be retried?

ramp-engine/ramp_engine/dispatcher.py
```diff
@@ -253,6 +259,10 @@ def update_database_results(self, session):
             update_all_user_leaderboards(session,
                                          self._ramp_config['event_name'])
 
+    @staticmethod
+    def _reset_single_submission(session, submission_id):
+        set_submission_state(session, submission_id, 'new')
```
rth, line 264: Maybe just directly call this set_submission_state in the dispatcher loop, without the added indirection of the added static function?

ramp-engine/ramp_engine/base.py
rth: No nevermind my previous comment, I just saw there was a `_is_submission_finished` so it does look like, `_is_submission_*` functions would map to a `worker._status` directly.

Then maybe a better name would be `_is_submission_interrupted` ? Because I would have assumed that `_is_submission_terminated` would map to a `killed` status and not a `restart` one.

We would also need to add the `restart` to the list possible statuses in the BaseWorker docstring.

rth (commented):
Thanks @lucyleeow ! I only reviewed the part outside of the aws scripts.

It might be useful to add a dispatcher test that a CondaEnvWorker with a `_status = 'retry' (or 'reset')` is indeed restarted, to test at least some of it locally?

lucyleeow:
Thanks for the review @rth. Suggestions added

> It might be useful to add a dispatcher test that a CondaEnvWorker with a _status = 'retry' (or 'reset') is indeed restarted, to test at least some of it locally?

I'm not good with RAMP and it's tests, and it doesn't seem to like me either. I tried adding a test (see recent commits) but I couldn't test locally because I got this error:

```txt
    @pytest.fixture(scope='session')
    def database_connection():
        """
        Create a Postgres database for the tests, and drop it when the tests are
        done.
        """
        config = safe_load(open("db_engine.yml"))
        dbowner = config.get('db_owner')
    
        engine = create_engine(f'postgresql://{dbowner}:@localhost/postgres',
                               isolation_level='AUTOCOMMIT')
    
        connection = engine.connect()
    
        database_config = read_config(database_config_template())
        username = database_config['sqlalchemy']['username']
        database_name = database_config['sqlalchemy']['database']
        try:
            connection.execute(f"""CREATE USER {username}
                                  WITH PASSWORD '{username}';
                                  ALTER USER {username} WITH SUPERUSER""")
        except exc.ProgrammingError as e:
>           raise ValueError(f'user {username} already exists') from e
E           ValueError: user mrramp already exists
```
but when I tried to delete mrramp with `ramp-database delete-user --login mrramp`, i got:
```
Traceback (most recent call last):
  File "/home/lucy/miniconda3/envs/ramp/bin/ramp-database", line 11, in <module>
    load_entry_point('ramp-database', 'console_scripts', 'ramp-database')()
  File "/home/lucy/Documents/RAMP/ramp-board/ramp-database/ramp_database/cli.py", line 365, in start
    main()
  File "/home/lucy/miniconda3/envs/ramp/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/lucy/miniconda3/envs/ramp/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/lucy/miniconda3/envs/ramp/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/lucy/miniconda3/envs/ramp/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/lucy/miniconda3/envs/ramp/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/lucy/Documents/RAMP/ramp-board/ramp-database/ramp_database/cli.py", line 59, in delete_user
    user_module.delete_user(session, login)
  File "/home/lucy/Documents/RAMP/ramp-board/ramp-database/ramp_database/tools/user.py", line 117, in delete_user
    user = session.query(User).filter(User.name == name).one()
  File "/home/lucy/miniconda3/envs/ramp/lib/python3.7/site-packages/sqlalchemy/orm/query.py", line 3289, in one
    raise orm_exc.NoResultFound("No row was found for one()")
sqlalchemy.orm.exc.NoResultFound: No row was found for one()
```

lucyleeow:
The tests seemed to have passed but I think the failure is due to time out?
```
================== 87 passed, 3 warnings in 116.35s (0:01:56) ==================
```
and
```
ramp-engine/ramp_engine/tests/test_aws.py ss                             [  7%]

ramp-engine/ramp_engine/tests/test_cli.py ......                         [ 28%]

ramp-engine/ramp_engine/tests/test_conda_worker.py ...........           [ 67%]

ramp-engine/ramp_engine/tests/test_daemon.py ..                          [ 75%]

ramp-engine/ramp_engine/tests/test_dispatcher.py ......

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

lucyleeow:
I've restarted the build....

rth:
> The frontend tests seemed to have passed but then I think it times out?

@lucyleeow I can reproduce locally, so it looks like it's a genuine issue with the test (that deadlocks at some point) rather than a CI issue. I'll try to have a look later today.

lucyleeow:
I haven't tested the test I added because https://github.com/paris-saclay-cds/ramp-board/pull/455#issuecomment-722753373
so that's likely to be the reason

rth:
Yeah, that error is strange..

lucyleeow:
It's probably a problem with the test - not know about ramp tests and not being able to test it means it's highly likely the test I wrote is incorrect...

lucyleeow:
commented out test in latest commit

maikia:
Hi @lucyleeow as for the test 
I think the problem is that if you call: `dispatcher._awaiting_worker_queue.get()` and there is nothing in that queue it will wait forever.

That said, you do put submissions in the queue when calling `dispatcher.launch()` but before you run `dispatcher._awaiting_worker_queue.get()` they are already processed and the queue is empty.

Non-optimal solution could be to create a submission manually right before getting worker from the queue:
```
    submissions = get_submissions(
        session_toy, event_config['ramp']['event_name'], 'training_error'
        )[0]
    submission_id, submission_name, _ = submissions

    # create the worker
    worker = dispatcher.worker(dispatcher._worker_config, submission_name)

    dispatcher._awaiting_worker_queue.put_nowait((worker, (submission_id,
                                                 submission_name)))
    worker, (_, submission_name) = dispatcher._awaiting_worker_queue.get()
```

But this still would not be sufficient because before you get to the worker status the submission is already killed.

Perhaps the submission would have to be set to something which takes longer to run? Is it possible within our test settings?

lucyleeow:
Also tried:
```
(ramp) lucy@:ramp-board(aws_spot)$ psql postgres lucy
psql (11.2)
Type "help" for help.

postgres=# DROP USER mrramp
postgres-# 
```
to remove user mrramp but did not work either.

lucyleeow:
Travis isn't building anymore

> Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.

lucyleeow:
This test works - tested locally. It may or may not test 
> It might be useful to add a dispatcher test that a CondaEnvWorker with a _status = 'retry' (or 'reset') is indeed restarted, to test at least some of it locally?

adequately.

PR #458, opened 2020-11-05
ADV/CLI Add --config-event parameter to compute_contributivity CLI

rth:
An improvement by @kegl to pass the `--event-config` argument, instead of the individual foder path in the compute compute_contributivity CLI. This is similar to what is done in other CLI endpoints.

Will merge to the `advanced` branch on green CI.

### DIFF
--- a/ramp-database/ramp_database/cli.py
+++ b/ramp-database/ramp_database/cli.py
@@ -7,7 +7,6 @@ import shutil
 
 from ramp_utils import read_config
 from ramp_utils import generate_ramp_config
-from ramp_utils import generate_worker_config
 
 from .utils import session_scope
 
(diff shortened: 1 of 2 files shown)


rth:
@kegl I made a few changes with respect to the initial version to be consistent with other CLI endpoints, for instance, [ramp-setup deploy-event](https://paris-saclay-cds.github.io/ramp-docs/ramp-board/stable/cli.html#ramp-setup-deploy-event),
 1. renamed `--event-config` to `--config-event` to be consistent
 2. removed `--event` (name) option, since this information is contained in the config, and writing test to make sure it's consistent with the config a bit more complex. Similarly other CLI endpoints take this informaiton from the config.

Feel free to merge if you are OK with this.

Issue #459, opened 2020-11-07
Travis CI credit limit reached


rth:
As mentioned by @lucyleeow in https://github.com/paris-saclay-cds/ramp-board/pull/455#issuecomment-723344504

> Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.

Also related scikit-learn thread: https://mail.python.org/pipermail/scikit-learn/2020-November/003900.html

That was fast, we used the initial credit allocation in just a few days.

What do you think we should do?  Ask for more credit (and if so how much?) or try to migrate to another CI (e.g. Github actions)?

cc @agramfort

agramfort:
I would move to github actions but it's some work :'(

>

PR #460, opened 2020-11-09
Add a github actions CI setup

rth:
Closes https://github.com/paris-saclay-cds/ramp-board/issues/459


rth:
Could someone please check that github actions are indeed enabled in the repo settings (including on PRs)? Not sure why they are not triggering. Thanks!

cc @agramfort @lucyleeow

rth:
Actions trigger on a test upstream branch https://github.com/paris-saclay-cds/ramp-board/tree/test-gh-actions, so they probably just need to be enabled to trigger on PRs from forks in the repo setting for this to work.

agramfort:
@rth you have now admin rights :)

lucyleeow:
@rth it might be because this PR is a draft: https://github.community/t/github-actions-not-being-triggered-on-draft-pr/16390

I couldn't figure out how to change it from a draft, maybe you have to do it yourself?

lucyleeow:
I think it's working now?

lucyleeow:
Doesn't seem to like python version?
```
Error: The value of "python-version" you provided was , which is invalid.
```

rth:
Thanks @lucyleeow and @agramfort !

I think this is good to go. All the same tests are run, and the coverage numbers are the same aside for 2 lines difference (that I think is due to codecov not being run for the last PR on master).

A nice thing is that we can have a separate flake8 job instead of doing that as part of tests (which is much faster).

One difference is that postgress is currently used from a separate docker image instead of the one from conda. That shouldn't matter. I'm actually not sure how the postgress process is run the background in TravisCI (I don't see the command anywhere).

Merging to unblock other PRs. But we can certainly improve on top of this.

lucyleeow:
Thanks @rth !

PR #462, opened 2020-11-12
MRG BUG enable use of on demand instances

maikia:
The use of on demand instances on AWS is no longer working.

This PR corrects it and adds a test

PR #464, opened 2020-11-13
Handle errors from AWS on creating instance

maikia:
Add handling of errors thrown by AWS when creating a spot instance
+ tests

### DIFF
--- a/ramp-engine/ramp_engine/aws/api.py
+++ b/ramp-engine/ramp_engine/aws/api.py
@@ -56,7 +56,7 @@ TRAIN_LOOP_INTERVAL_SECS_FIELD = 'train_loop_interval_secs'
 MEMORY_PROFILING_FIELD = 'memory_profiling'
 
 # how long to wait for connections
-WAIT_MINUTES = 10
+WAIT_MINUTES = 0.03
 MAX_TRIES_TO_CONNECT = 6
 
 HOOKS_SECTION = 'hooks'
--- a/ramp-engine/ramp_engine/aws/api.py
+++ b/ramp-engine/ramp_engine/aws/api.py
@@ -172,6 +172,7 @@ def launch_ec2_instances(config, nb=1):
                     ValidFrom=now,
                     ValidUntil=(now + request_wait),
                 )
+                break
             except botocore.exceptions.ClientError as e:
                 n_try += 1
                 if n_try < max_tries_to_connect:
--- a/ramp-engine/ramp_engine/aws/api.py
+++ b/ramp-engine/ramp_engine/aws/api.py
@@ -55,6 +55,10 @@ LOCAL_LOG_FOLDER_FIELD = 'logs_dir'
 TRAIN_LOOP_INTERVAL_SECS_FIELD = 'train_loop_interval_secs'
 MEMORY_PROFILING_FIELD = 'memory_profiling'
 
+# how long to wait for connections
+WAIT_MINUTES = 10
+MAX_TRIES_TO_CONNECT = 6
+
 HOOKS_SECTION = 'hooks'
 HOOK_START_TRAINING = 'start_training'
 HOOK_SUCCESSFUL_TRAINING = 'successful_training'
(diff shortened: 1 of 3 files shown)


maikia:
I still need to handle/log some errors from the dispatcher level but if you could already review this one it would move things forward a bit faster

thanks in advance
@agramfort @lucyleeow @glemaitre @tomMoral

ramp-engine/ramp_engine/aws/api.py
```diff
@@ -148,21 +152,43 @@ def launch_ec2_instances(config, nb=1):
...
+                                   f' {n_try} out of {max_tries_to_connect}'
+                                   ' tries to connect)')
+                    time.sleep(wait_minutes*60)
+                else:
+                    logger.error(f'Not enough instances available: {e}')
+                    return None,
+            except Exception as e:
+                # unknown error
+                logger.error(f'AWS worker error: {e}')
+                return None,
```
tomMoral, line 191: Not a big fan of `while` loop because I feel it is harder to read what is hapening.
I propose to use `for/break/else` structure to make if a bit clearer but this is mainly nitpicking so feel free to ignore the comment if it does not seem clearer to you:

```suggestion
        for n_try in range(max_tries_to_connect):
            try:
                response = client.request_spot_instances(
                    AvailabilityZoneGroup=config[REGION_NAME_FIELD],
                    InstanceCount=nb,
                    LaunchSpecification={
                        'SecurityGroups': [security_group],
                        'ImageId': ami_image_id,
                        'InstanceType': instance_type,
                        'KeyName': key_name,
                    },
                    Type='one-time',
                    ValidFrom=now,
                    ValidUntil=(now + request_wait),
                )
                break
            except botocore.exceptions.ClientError as e:
                # wait before you try again
                logger.warning('Not enough instances available: I am going'
                               f' to wait for {wait_minutes} minutes'
                               ' before trying again (this was'
                               f' {n_try} out of {max_tries_to_connect}'
                               ' tries to connect)')
                time.sleep(wait_minutes*60)
            except Exception as e:
                # unknown error
                logger.error(f'AWS worker error: {e}')
                return None,
        else:
            # reached max retries
            logger.error(f'Not enough instances available: {e}')
            return None,
```

ramp-engine/ramp_engine/aws/api.py
```diff
@@ -55,6 +55,10 @@
 TRAIN_LOOP_INTERVAL_SECS_FIELD = 'train_loop_interval_secs'
 MEMORY_PROFILING_FIELD = 'memory_profiling'
 
+# how long to wait for connections
+WAIT_MINUTES = 10
+MAX_TRIES_TO_CONNECT = 6
```
tomMoral, line 60: 1hour waiting time seems a lot no?
But I have no idea how frequent the failures are?

tomMoral (commented):
I don't know the code enough to tell you if the test is good but it seems nice to have a retry mechanism.

Just a small comment on the while loop that I suggest to replace with a `for` loop to get more readable code but it might be personal taste :)

ramp-engine/ramp_engine/aws/api.py
maikia: Thanks for your comments :-) @tomMoral
Yes. 1h seems like a lot but I think in my particular case (when there are only 3 instances available) It's good to give the submission a bit of a chance to get submitted: otherwise it will fail and the error will be coming from not having instances, not the wrong submission.
I think it would be better to put those two params in the config file of the event in the future, but for now I prefer to keep it as is so the events (especially the locomotion challenge) would not have to be redeployed

lucyleeow:
Can you give a log of the errors from creating AWS instance? If the problem is that the dispatcher thinks we have more instances than we actually have - maybe the problem would be better fixed by correcting that and not waiting so long between requests...

ramp-engine/ramp_engine/aws/api.py
lucyleeow: > when there are only 3 instances available)

It might be worth following up with Nicolas about this. Also requesting for increase in spot instance limit seems to be much easier than requesting for increase in on-demand instance, I think if you requested to increase spot instance limit it should go through

ramp-engine/ramp_engine/aws/worker.py
```diff
@@ -61,13 +61,16 @@ def setup(self):
...
             self.submission))
         self.instance, = aws.launch_ec2_instances(self.config)
+
         if self.instance:
             logger.info("Instance launched for submission '{}'".format(
                 self.submission))
         else:
-            logger.info("Unable to launch instance for submission "
-                        "'{}'".format(self.submission))
+            logger.error("Unable to launch instance for submission "
```
lucyleeow, line 69: nitpick, change this to f-string?

maikia:
the error is this one: botocore.exceptions.ClientError thrown from Amazon and not caught by the dispatcher nor by the worker

@lucyleeow I agree that keeping the dispatcher waiting for just a single worker would not be a good idea. we can shorten the time of waiting for now

lucyleeow:
>  botocore.exceptions.ClientError

Is there an error message with that?

lucyleeow:
> AWS service exceptions are caught with the underlying botocore exception, ClientError. After you catch this exception, you can parse through the response for specifics around that error, including the service-specific exception. Exceptions and errors from AWS services vary widely. You can quickly get a list of an AWS service’s exceptions using Boto3.

sounds like this is a relatively generic error and the message gives more details?

maikia:
`botocore.exceptions.ClientError: An error occurred (MaxSpotInstanceCountExceeded) when calling the RequestSpotInstances operation: Max spot instance count exceeded`

ramp-engine/ramp_engine/aws/api.py
maikia: I requested in spot instance limit -> the use of on demand is now broken on the current release of RAMP anyways
(I made the PR to fix it: https://github.com/paris-saclay-cds/ramp-board/pull/462).

agramfort (approved):
tests are green so it's good from my end. Let's try and see what it gives. If we break things it's no big deal. We can fix / revert quickly as long as we do not touch the DB

lucyleeow:
> tests are green so it's good from my end. 

@agramfort FYI I think aws tests are skipped on CI

agramfort:
arfff :(

any chance to fix this? using free micro spot instances?

>

rth:
FYI, there is https://github.com/spulec/moto for mocking AWS services in tests (though I have no experience using it)

maikia:
There are two aws tests which are skipped. There are virtually no other tests except what I have added here <- which should not have been skipped by CI unless I am not looking at the right settings (?)

I am merging

Thanks @tomMoral @rth @lucyleeow and @agramfort

PR #465, opened 2020-11-13
Dispatcher: handle worker errors

maikia:
The dispatcher gets some unhandled errors when running the worker and it breaks
this is to ensure that the dispatcher does not break


agramfort (approved):

rth (approved):
This is great, thank you @maikia !

I also run into issues of dispatcher stopping due to unhandled exception in the past.

PR #467, opened 2020-11-16
Add timout to ssh

maikia:
Adds a timeout to curl and handles the exceptions in case there are any errors with subprocess output (before the connection just hanged sometimes stopping the work of the whole dispatcher).

The problem which remains and I don't know how to approach is that even if the connection is not established after (what is set now) 9 tries, with initial time 1sec doubled after each next try (ie max waiting time 8.51 min) the only thing we are able to return is either that the instance finished or not (ie True or False). If for any reason the instance is lost (shutting by hand, other scenarios (??) ) it will still try connecting to it the next time again.

This might be a problem for another PR though..

### DIFF
--- a/ramp-engine/ramp_engine/aws/api.py
+++ b/ramp-engine/ramp_engine/aws/api.py
@@ -940,17 +940,20 @@ def is_spot_terminated(config, instance_id):
     'instance-action' will be present."""
     cmd_timeout = 1
     n_retry = 5
-    #cmd = "curl http://169.254.169.254/latest/meta-data/instance-action" # -m cmd_timeout --retry n_retry
-    #try:
-    out = _run(config, instance_id, cmd, return_output=True)
-    # except subprocess.CalledProcessError:
-    #     logger.error('Unable to run curl: {e}')
-    #    terminated = False
-    #except Exception as e:
-    #    logger.error('Unhandled exception happend when checking if there'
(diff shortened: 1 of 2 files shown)


ramp-engine/ramp_engine/aws/api.py
```diff
@@ -938,9 +938,22 @@ def is_spot_terminated(config, instance_id):
...
+
+    try:
+        out = _run(config, instance_id, cmd, return_output=True)
+        out = out.decode('utf-8')
+    except subprocess.CalledProcessError:
+        logger.error('Unable to run curl: {e}')
+        return False
+    except Exception as e:
+        logger.error('Unhandled exception happend when checking if there'
+                     f' is an instance action: {e}')
```
lucyleeow, line 954:
```suggestion
        logger.error('Unhandled exception occurred when checking for '
                     f'instance action: {e}')
```

lucyleeow (commented):
Any idea why the connection to AWS just hangs some times? If it hangs here, maybe it will for other subsequent ssh requests we send to?

maikia:
thanks for your review @lucyleeow 

Yes. I spent quite some time investigating the whole hanging issue. It seems that curl was hanging for now, but there is always a possibility it will hang elsewhere as well. 

What other places would you suggest we need to take care of before making release?

maikia:
Thanks @lucyleeow

PR #469, opened 2020-11-18
BUG error in the name of the submission state

maikia:
There is a mistake in the name of the submission ('send_to_training' instead of 'sent_to_training') so each time the dispatcher is restarted the submissions with the status 'sent_to_training' hang


agramfort (approved):

rth:
Thanks @maikia ! Issue introduced in https://github.com/paris-saclay-cds/ramp-board/pull/360. Indeed I also experienced strange behavior when restarting.

PR #470, opened 2020-11-19
BUG Catch ssh error

maikia:
It still happens that the dispatcher breaks on the the ssh connection. This is usually on the ssh with rsync command.

This PR is an attempt to fix this in 
- download_log
- download_prediction
- upload_submission


ramp-engine/ramp_engine/tests/_config.yml
```diff
@@ -5,9 +5,6 @@ ramp:
     event_is_public: true
     data_dir: ramp-board/ramp_deployment/ramp-data/iris
     kit_dir: ramp-board/ramp-engine/ramp_engine/tests/kits
-    submissions_dir: ramp-board/ramp-engine/ramp_engine/tests/kits/iris/submissions
-    predictions_dir: ramp-board/ramp-engine/ramp_engine/tests/kits/iris/predictions
-    logs_dir: ramp-board/ramp-engine/ramp_engine/tests/kits/iris/logs
```
rth, line 10: The submission dir should be in the ramp section shoudn't it?

maikia: @rth 
not really. Here are the docs on how the config for AWS should look like:
https://paris-saclay-cds.github.io/ramp-docs/ramp-board/stable/workers.html#event-configuration

The worker gets the submissions from the AWS and needs to know which directory to copy them to on the server:
https://github.com/paris-saclay-cds/ramp-board/blob/fa193ace47a1c14c7d80d0b999704df8f6018923/ramp-engine/ramp_engine/aws/worker.py#L76

ramp-engine/ramp_engine/aws/worker.py
```diff
@@ -135,19 +135,37 @@ def collect_results(self):
...
-        else:
-            error_msg = _get_traceback(
-                aws._get_log_content(self.config, self.submission))
-            self.status = 'collected'
+        exit_status = 0
+        try:
+            _ = aws.download_log(self.config,
+                                 self.instance.id, self.submission)
+        except Exception as e:
+            logger.error("Error happend when downloading the logs"
```
lucyleeow, line 143:
```suggestion
            logger.error("Error occurred when downloading the logs"
```

ramp-engine/ramp_engine/tests/_config.yml
rth: Ahh, interesting! This does solve an issue I was having in DaskWorker about having a separate local and remote predictions_dir. I should read more on how AWS worker works. But then isn't it incompatible with the [`generate_worker_config`](https://github.com/paris-saclay-cds/ramp-board/blob/fa193ace47a1c14c7d80d0b999704df8f6018923/ramp-utils/ramp_utils/worker.py#L46) util which explicitly loads this information from the `ramp` config section? 

Fine to merge this but, maybe we could rename this file to `_aws_config.yml` or something more specific (not necessarily in this PR) ?

maikia: Hm. you are right. the submission folder is not a mandatory parameter in the ramp config and if it is not given the default will be used:
https://github.com/paris-saclay-cds/ramp-board/blob/fa193ace47a1c14c7d80d0b999704df8f6018923/ramp-utils/ramp_utils/ramp.py#L92

but this will probably cause problems if the worker receives different submission directory from what is in  ramp_config.

I am not sure what would be the best way out of this

rth: Specifying it in both sections (one for remote and one for local)?

maikia: you mean requiring the user to type the same thing in the ramp_event config in the ramp section and in the worker section?

maikia:
@rth are you ok if I merge this one?
It's not really related to the section locations anyway, just catching bugs...

rth:
Yes, sure no objections. Thanks @maikia !

maikia:
thanks @rth and @lucyleeow

PR #472, opened 2020-11-24
ENH gets latest image if more than one

maikia:
**use case**: when pipeline is used each next version of the image will have the name image_name_{date_time_of_creation}

this PR enables a worker to find all the images which include the name image_name in their name and to select the newest one.

This only works if the image_name is given in the config file. If the id of the image will be given the worker will always use the same image with that ID.

### DIFF
--- a/ramp-engine/ramp_engine/aws/api.py
+++ b/ramp-engine/ramp_engine/aws/api.py
@@ -251,26 +251,28 @@ def launch_ec2_instances(config, nb=1):
 def _get_image_id(config, image_name):
     sess = _get_boto_session(config)
     client = sess.client('ec2')
+
+    # get all the images with the given imange_name in the name
     result = client.describe_images(Filters=[
         {
             'Name': 'name',
-            'Values': [
-                image_name
-            ]
+            'Values': [image_name+'*'
+                       ],
         }
     ])
+
     images = result['Images']
     if len(images) == 0:
         raise ValueError(
             'No image corresponding to the name "{}"'.format(image_name))
     elif len(images) > 1:
-        raise ValueError(
-            'Multiple images corresponding to the name "{}".'
-            ' Please fix that'.format(image_name))
+        # get only the newest image if there are more than one
+        image = sorted(images, key=lambda x: x['CreationDate'],
+                       reverse=True)[0]
     else:
         image = images[0]
-        image_id = image['ImageId']
-        return image_id
+    image_id = image['ImageId']
+    return image_id
 
 
 def terminate_ec2_instance(config, instance_id):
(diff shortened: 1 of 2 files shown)


maikia:
not sure how to make tests for it though....

agramfort:
merge if you need to move on. We will know very quickly if it works :)

ramp-engine/ramp_engine/aws/api.py
```diff
@@ -251,26 +251,28 @@ def launch_ec2_instances(config, nb=1):
...
             'No image corresponding to the name "{}"'.format(image_name))
     elif len(images) > 1:
-        raise ValueError(
-            'Multiple images corresponding to the name "{}".'
-            ' Please fix that'.format(image_name))
+        # get only the newest image if there are more than one
+        image = sorted(images, key=lambda x: x['CreationDate'],
+                       reverse=True)[0]
     else:
         image = images[0]
```
tomMoral, line 273: you could actually remove the `elif` and the `else`.
It will work so let's not complexify the code here.

ramp-engine/ramp_engine/aws/api.py
```diff
@@ -251,26 +251,28 @@ def launch_ec2_instances(config, nb=1):
...
+
+    # get all the images with the given imange_name in the name
     result = client.describe_images(Filters=[
         {
             'Name': 'name',
-            'Values': [
-                image_name
-            ]
+            'Values': [image_name+'*'
+                       ],
```
tomMoral, line 260: This is quite efficient :+1: 
Just a small nitpick
```suggestion
            'Values': [
                f'{image_name}*'
            ],
```

tomMoral (approved):
Some small nitpicks but otherwise, looks good to me.

ramp-engine/ramp_engine/aws/api.py
```diff
@@ -251,26 +251,26 @@ def launch_ec2_instances(config, nb=1):
 def _get_image_id(config, image_name):
     sess = _get_boto_session(config)
     client = sess.client('ec2')
+
+    # get all the images with the given imange_name in the name
```
tomMoral, line 255:
```suggestion
    # get all the images with the given image_name in the name
```

ramp-engine/ramp_engine/aws/api.py
```diff
@@ -251,26 +251,26 @@ def launch_ec2_instances(config, nb=1):
...
-    else:
-        image = images[0]
-        image_id = image['ImageId']
-        return image_id
+
+    # get only the newest image if there are more than one
+    image = sorted(images, key=lambda x: x['CreationDate'],
+                   reverse=True)[0]
+    image_id = image['ImageId']
+    return image_id
```
tomMoral, line 273:
```suggestion
    return image['ImageId']
```

tomMoral (approved):
Two nitpicks but good to go otherwise :)

tomMoral (approved):
LGTM!

maikia:
Thanks @tomMoral and @agramfort

Issue #471, opened 2020-11-23
BUG creation of AWS ghost instances


maikia:
Sometimes ghost AWS instances are created: ie instances which are not assigned any submission number (from the perspective of AWS) and which just hang with a similar profile of CPU utilization:

![image](https://user-images.githubusercontent.com/2219642/99942346-d4382080-2d6f-11eb-8d48-596b66483bdd.png)

Also not much bytes are send in or out.

Could be that on some errors RAMP creates an instance but fails to send the submission and it fails to terminate that instance.

lucyleeow:
I think we tag the instance with the submission name with `_tag_instance_by_submission` at the end of `launch_train` - potentially it hangs during a step in `launch_train` before `_tag_instance_by_submission`? Seems possible as we do the training here:
```Python
def launch_train(config, instance_id, submission_name):
    """
    Launch the training of a submission on an ec2 instance.
    A screen named `submission_folder_name` (see below)
    is created and within that screen `ramp_test_submission`
    is launched.

    Parameters
    ----------

    instance_id : str
        instance id

    submission_id : int
        submission id
    """
    ramp_kit_folder = config[REMOTE_RAMP_KIT_FOLDER_FIELD]
    values = {
        'ramp_kit_folder': ramp_kit_folder,
        'submission': submission_name,
        'submission_folder': os.path.join(ramp_kit_folder, SUBMISSIONS_FOLDER,
                                          submission_name),
        'log': os.path.join(ramp_kit_folder, SUBMISSIONS_FOLDER,
                            submission_name, 'log')
    }
    # we use python -u so that standard input/output are flushed
    # and thus we can retrieve the log file live during training
    # without waiting for the process to finish.
    # We use an espace character around "$" because it is interpreted
    # before being run remotely and leads to an empty string
    run_cmd = (r"python -u \$(which ramp_test_submission) "
               r"--submission {submission} --save-y-preds ")
    if config.get(MEMORY_PROFILING_FIELD):
        run_cmd = (
            "mprof run --output={submission_folder}/mprof.dat "
            "--include-children " + run_cmd)
    cmd = (
        "screen -dm -S {submission} sh -c '. ~/.profile;"
        "cd {ramp_kit_folder};"
        "rm -fr {submission_folder}/training_output;"
        "rm -f {submission_folder}/log;"
        "rm -f {submission_folder}/mprof.dat;"
        + run_cmd + ">{log} 2>&1'"
    )
[...]

PR #473, opened 2020-11-24
catch possible unnknown errors when loading submission

maikia:
It catches possible unknown errors when training the submission

hopefully closes: #471


ramp-engine/ramp_engine/tests/test_aws.py
```diff
@@ -199,6 +199,33 @@ def test_launch_ec2_instances(boto_session_cls, use_spot_instance):
     launch_ec2_instances(config['worker'])
 
 
+@mock.patch('ramp_engine.aws.api._rsync')
+@mock.patch('ramp_engine.aws.api.launch_train')
+def test_aws_worker_launch_train_error(launch_train, test_rsync, caplog):
```
rth, line 204: Nice!

rth (approved):
Thanks @maikia ! LGTM though I'm not too closely acquainted with the AWS worker code base.

lucyleeow (commented):
Looks good. Is the instance killed when worker status is 'error' ? This may be related to #474, as when the work status is error, we do this:
```Python
            if worker.status == 'error':
                set_submission_state(session, submission_id, 'checking_error')
                continue
```
which I don't think kills the instance so we have the 'ghost instances'.

maikia:
> Looks good. Is the instance killed when worker status is 'error' ? This may be related to #474, as when the work status is error, we do this:
> 
> ```python
>             if worker.status == 'error':
>                 set_submission_state(session, submission_id, 'checking_error')
>                 continue
> ```
> 
> which I don't think kills the instance so we have the 'ghost instances'.

good point @lucyleeow 
What do you think now?

lucyleeow:
This is good - with regards to the `teardown` when status is error, why did we not have this already for the condaworkers? I am not so familiar with the condaworkers - I am just wondering if there is/was any reason to not `teardown` in this place, before?

I also wonder if we also need to add a worker 'error status and teardown to the `collect_result` results part of the dispatcher. During this part the dispatcher is continuously checking if the worker has finished via `_wait_until_train_finished`, `_training_finished` and `_has_screen`. It seems possible to get error here as well - as you pointed out in #479

maikia:
yes. You are right @lucyleeow . I was also wondering why teardown was not called here:
https://github.com/paris-saclay-cds/ramp-board/blob/26762a460d188033aea29144471b52aa06062727/ramp-engine/ramp_engine/dispatcher.py#L227

I couldn't find where it was called elsewhere when error status is set. 
Maybe here? (although it does not seem to answer it fully): https://github.com/paris-saclay-cds/ramp-board/blob/26762a460d188033aea29144471b52aa06062727/ramp-engine/ramp_engine/local.py#L126 

@glemaitre @rth could please confirm?

glemaitre:
It depends if you want to do the action of the teardown which usually clean-up stuff.
If you just want to kill the worker but still be able to access the data that was computed, you probably don't want to call the teardown.

maikia:
Thanks @glemaitre 
So do you agree on putting teardown here: for all the checking errors: https://github.com/paris-saclay-cds/ramp-board/blob/26762a460d188033aea29144471b52aa06062727/ramp-engine/ramp_engine/dispatcher.py#L171 
which are usually unknown errors coming from the worker itself.
Perhaps it is not an ideal long term solution and we should teardown or not depending on the exact error, but for now we are left with 'ghost instances' on AWS and I think this should solve the problem. But perhaps it could introduce another one if we are loosing the data on teardown?

ramp-engine/ramp_engine/dispatcher.py
```diff
@@ -168,6 +168,11 @@ def launch_workers(self, session):
                 worker.status = 'error'
             if worker.status == 'error':
                 set_submission_state(session, submission_id, 'checking_error')
+                worker.teardown()  # kill the worker
+                self._logger.info(
+                            f'Worker {worker} killed due to an error '
+                            f'during training'
+                        )
```
glemaitre, line 175:
```suggestion
                self._logger.info(
                    f'Worker {worker} killed due to an error '
                    f'during training'
                )
```

maikia:
Thanks @glemaitre @lucyleeow and @rth

PR #476, opened 2020-11-25
Add DaskWorker to master (#452)

rth:
Backport of https://github.com/paris-saclay-cds/ramp-board/pull/452 to the master branch.

This worker type was used for an actual event with no major issues. The goal of having it in master is to reduce divergence between master and advanced branches, and avoid for instance that fixes to the aws worker inadvertently break the dask worker. Running DaskWorker tests (included in this PR) on master would prevent that.

This should have impact on other workers.


rth:
OK, I'll take the responsibility of merging this. The original PR was reviewed by @albertcthomas in https://github.com/paris-saclay-cds/ramp-board/pull/452#pullrequestreview-523364899.

This should not affect existing workers, and I'll double check that this doesn't impact the public RAMP server. This accounts for a good amount of the diff from the advanced branch, and is a step forward for merging those branches together. The tiny decrease in coverage,

>  93.66% (-0.02%) compared to 68a5ec8 

is explained by the few lines that cannot be tested without a remote dask cluster. The code coverage for the tests with a local dask cluster (done in CI) should be quite good otherwise, most of the tests from the conda worker are re-used. We have also used this dask worker for an actual event without much issues.

kegl:
Good to merge for me, I trust you on this, thx @rth

PR #484, opened 2020-11-30
Retry not wait when connecting to the instance on submiting solution

maikia:
Sometimes the worker cannot connect to the instance because there are not enough instances are available. This can happen for multiple of reasons one of them being that the instance which was set to terminate do not yet is available for the use.
For that reason we previously added a possibility to wait and try again few times before giving an error. 

This was not very efficient because it forced the whole dispatcher to wait this time and not allowing it to collect the results from other workers and possibly free an instance -> which would solve the problem of not having enough instances if the cause was different from the above.

This PR shortens the waiting time and if the instances are still not available it puts the worker back into the queue to try later.

### DIFF
--- a/ramp-engine/ramp_engine/aws/api.py
+++ b/ramp-engine/ramp_engine/aws/api.py
@@ -57,7 +57,7 @@ MEMORY_PROFILING_FIELD = 'memory_profiling'
 
 # how long to wait for connections
 WAIT_MINUTES = 2
-MAX_TRIES_TO_CONNECT = 5
+MAX_TRIES_TO_CONNECT = 1
 
 HOOKS_SECTION = 'hooks'
 HOOK_START_TRAINING = 'start_training'
@@ -185,11 +185,11 @@ def launch_ec2_instances(config, nb=1):
                     time.sleep(wait_minutes*60)
                 else:
                     logger.error(f'Not enough instances available: {e}')
-                    return None,
+                    return None, 'retry'
             except Exception as e:
                 # unknown error
                 logger.error(f'AWS worker error: {e}')
-                return None,
+                return None, e
         # Wait until request fulfilled
         waiter = client.get_waiter('spot_instance_request_fulfilled')
         request_id = \
@@ -242,10 +242,10 @@ def launch_ec2_instances(config, nb=1):
         waiter = client.get_waiter('instance_status_ok')
         try:
             waiter.wait(InstanceIds=instance_ids)
-        except botocore.exceptions.WaiterError:
-            return None,
+        except botocore.exceptions.WaiterError as e:
+            return None, e
 
-    return instances
+    return instances, 0
 
 
 def _get_image_id(config, image_name):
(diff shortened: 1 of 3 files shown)


agramfort:
good to go from your end?

maikia:
I think so. but it's always better if someone reviews it...

ramp-engine/ramp_engine/aws/api.py
```diff
@@ -57,7 +57,7 @@
 
 # how long to wait for connections
 WAIT_MINUTES = 2
-MAX_TRIES_TO_CONNECT = 5
+MAX_TRIES_TO_CONNECT = 1
```
lucyleeow, line 60: If we change this to 1, do we still need the `if n_try < max_tries_to_connect:` ? I guess leaving it gives us the option to increase `MAX_TRIES_TO_CONNECT` in future ?

lucyleeow (commented):
Only giving it quick look - sorry i couldn't find the part where it is added to the queue? Can you point me in the right direction? Thanks!

ramp-engine/ramp_engine/aws/api.py
maikia: Yes. I prefer to leave it as a param for two reasons:
1. it is easier to test because for the test we can set this and the `WAIT_MINUTES` to low values so that the tests is done within reasonable time
2. we can easily update it in the future

maikia:
> Only giving it quick look - sorry i couldn't find the part where it is added to the queue? Can you point me in the right direction? Thanks!

When the worker status is set to 'retry' dispatcher sets the submission back to `new`, ie puts it back to the queue

https://github.com/paris-saclay-cds/ramp-board/blob/75e265549a4205fe3ef144cd0cb5477e890cbfb8/ramp-engine/ramp_engine/dispatcher.py#L219

maikia:
@agramfort 
We can test locally, but if you want to have a staging server we can have a continuously running EC2 instance that we can configure ramp on. We could even provision the database with a dump from the ramp.studio. 


> Only giving it quick look - sorry i couldn't find the part where it is added to the queue? Can you point me in the right direction? Thanks!

When the worker status is set to 'retry' dispatcher sets the submission back as `new`

https://github.com/paris-saclay-cds/ramp-board/blob/75e265549a4205fe3ef144cd0cb5477e890cbfb8

maikia:
@agramfort @lucyleeow if you are happy could you pls merge (then it could already be present in the next release)

ramp-engine/ramp_engine/aws/api.py
lucyleeow: > it is easier to test because for the test we can set this and the WAIT_MINUTES to low values so that the tests is done within reasonable time

you mean `MAX_TRIES_TO_CONNECT` ? But yes if you want to use several tries in future, though adding it to queue might be better

lucyleeow:
Ah I was looking for an explicit adding to the queue but I had forgotten I added the 'retry' function! LGTM

agramfort:
thx @maikia !

agramfort:
for the staging it's up to you if it helps you or not

PR #489, opened 2020-12-07
DOC use private instead of public data

maikia:
Corrects a mistake in the docs suggesting using public instead of private dataset on the AWS instances.


rth (approved):
Thanks!

PR #506, opened 2021-01-11
ENH AWS remove availibity zone group from the setting up instance

maikia:
by setting availability zone group we are limiting our possible resources, therefore it is better not to set it.


agramfort:
REGION_NAME_FIELD is still used somewhere?

maikia:
Yes:
https://github.com/paris-saclay-cds/ramp-board/blob/b2bd7e398c27a9a3842d087c301ba07a393f46e4/ramp-engine/ramp_engine/aws/api.py#L897

agramfort:
Thx

PR #512, opened 2021-04-09
MAINT Add missing pip dependency in ci_tools/environment_iris_kit.yml

rth:
I have recently seen some CI failures in https://github.com/paris-saclay-cds/ramp-board/pull/476 which should be unrelated to that PR, so opening a separate PR to investigate.

### DIFF
--- a/ci_tools/environment_iris_kit.yml
+++ b/ci_tools/environment_iris_kit.yml
@@ -5,5 +5,6 @@ dependencies:
   - numpy
   - scikit-learn
   - pandas
+  - pip
   - pip:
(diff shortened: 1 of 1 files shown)


rth:
OK, in the end it was a transient issue with OpenML https://github.com/openml/OpenML/issues/1092 while downloading iris data from there (https://github.com/ramp-data/iris/issues/1).

This PR now contains only a very minor fix to avoid warning about missing pip dependency when creating the conda env.

kegl:
Thanks a lot @rth!

Issue #516, opened 2021-04-30
Server backup procedure


rth:
So think there are 3 main points necessairy to perform a server backup, 

## Full backup

 1. the PostgreSQL DB, probably with doing an [SQL Dump](https://www.postgresql.org/docs/9.1/backup-dump.html). Currently ~121GB on the main server (but likely many duplicate files). 
 2. the ramp_deployment folder (including the submissions). Currently 250GB on the main server. Submissions might not be strictly necessary. 
 3. conda envs to facilitate re-deployment. Currently 24GB.

These backups can then be copied into some cloud storage, for instance with [rclone](https://rclone.org/) 

Ideally with encryption, particularly for the DB.

## Full backup

The last two points can probably be done incrementally. Less sure about the first one.

agramfort:
sounds good !

>

PR #120, opened 2019-01-16
MAINT: cleanup old files

glemaitre:

PR #521, opened 2021-05-12
DOC Database migrations with flask-migrate & alembic

rth:
This adds documentation on how to perform database migrations with [flask-migrate](https://github.com/miguelgrinberg/Flask-Migrate) (based on [alembic](https://pypi.org/project/alembic/)).

The actual alembic migration files should be committed to an instance dependent repository. For instance for https://ramp.studio that would be https://github.com/agramfort/ramp-studio-maintenance (private). As far as I can tell there used to be such `migration/` folder previously as part of this repository https://github.com/paris-saclay-cds/ramp-board/pull/120

I'm not sure: has anyone done migrations on the ramp.studio server in the last few years cc @maikia @glemaitre @agramfort? Did you also used flask-migrate or alembic directly?

I can't find any mentions that leaving that `Migrate` call in the app definition is counter-indicated, but otherwise we could also put it behind some env variable.

The end goal with more systematic backups #516 is allow making DB migrations less exceptional (and risky). 

*Note:* the CI might fail due to a new flask 2.0 release and because of unpinned dependencies. Will look into that separately.

### DIFF
--- a/doc/server-maintenance.rst
+++ b/doc/server-maintenance.rst
@@ -1,0 +1,26 @@
+##################
+Server maintenance
+##################
+
+Database migrations
+-------------------
+
+Please follow the
+`flask-migrate <https://flask-migrate.readthedocs.io/en/latest/>`_ documentation
+to perform database migrations.
+
+Note that the corresponding migration files **should not** be added to the
+ramp-board repository but rather to a separate per-instance git repository.
+
+You may need to add the following file to this repository,
+
+.. code:: python
+
+    from ramp_frontend.wsgi import make_app
+
+    app = make_app("<path to the config.yaml>")
+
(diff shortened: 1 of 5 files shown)


glemaitre:
I did it when we migrated the previous database and I think I used Alembic.
I remember to have kept the old files related to migration in a legacy repository: https://github.com/paris-saclay-cds/ramp-board-legacy/tree/master/migrations

rth:
Thanks for the input @glemaitre . Actually yes, just having a separate `app.py` for migrations as was done in that repo is better. This means that this becomes a purely documentation PR with no code changes.

agramfort (approved):

agramfort:
thx @rth

Issue #511, opened 2021-02-05
Ability to delete a user account in the UI


rth:
It would be good to have a UI button to permanently delete a user account, including all personal data, and code submissions. 

This is required by GDPR https://github.com/paris-saclay-cds/ramp-board/issues/440

Issue #440, opened 2020-08-19
Better GDPR compliance


rth:
While looking into preparation of a workshop with @kegl , it made me realize that a number of things regarding GDPR, handling of personal data and informing users could probably be improved, even for the https://ramp.studio/ instance.

In particular,
 - [ ] adding an optional template for the privacy policy explaining how their personal data (mostly name, email and submitted code) is stored and processed (which should be customized for deployed instances). Also clarifying the copyright situation with the submitted code (i.e. that users keep the copyright).
 - [ ] when users sign-up having a check-box that that agree with the privacy policy and likely more specific actions their email is used for
 - [ ] add UI to remove an account, which should also remove all the submitted data
 - [ ] add information about usage of cookies. Currently one is used by flask-login with a storage policy of 1 year (used to remember credentials). This can probably  be reduced to a few months by default.
 - [ ] likely discuss the the data retention policy. The easiest here might be that events after X years are removed, though say Kaggle doesn't do it, so I'm not sure how they get around it.

agramfort:
+1000 !

>

PR #524, opened 2021-05-16
ENH UI to delete account

rth:
This adds a UI so that users can delete their account.

Closes https://github.com/paris-saclay-cds/ramp-board/issues/511 and helps for the GDPR compliance https://github.com/paris-saclay-cds/ramp-board/issues/440

Under "Update profile" there is now a "Delete account" button,

image

Which when clicked triggers a confirmation form,
image

And after clicking OK all personal data will be removed. The existing user entry will be re-initialized to default values (i.e. empty fields) except for the following field,
 - `firstname`: "deleted"
 - `lastname`: "deleted"
 - `email`: "{user.id}@deleted.com"
 - `hashed_password`: randomly re-generated, and not the output of the hash function so no user input will ever match.

The 1 person team name will also be renamed to "deleted_{user.id}", however leaderboard will not be re-computed so the previous name would still be shown there.

Submissions will however not be removed. This similar to what happens when a Github account is removed (i.e. contributed code is not removed) and assumes that there are Terms of Use that mention that code essentially donated to the RAMP platform. Anyway removing all submissions for a users is technically more complex (need to re-compute leaderboards etc). Also for audit purposes it's still useful to have some records that the user existed.

Note that `access_level` remains unchanged, which makes it a bit less natural to filter deleted accounts. This is because, 
 - currently this works without a DB migration, and if we add `deleted` to the [`access_level` Enum](https://github.com/paris-saclay-cds/ramp-board/blob/675833488eb8c7f92f98931e7726d4483dface18/ramp-database/ramp_database/model/user.py#L131) it will require a migration, and besides this particular migration is not supported by alembic https://github.com/sqlalchemy/alembic/issues/278 requiring manual workarounds. 
 - also having that extra information about the account state can be useful.

### DIFF
--- a/ramp-frontend/ramp_frontend/templates/update_profile.html
+++ b/ramp-frontend/ramp_frontend/templates/update_profile.html
@@ -132,14 +132,38 @@
         <input type="submit" value="Update" class="ui submit button">
     </div>
     </form>
-  </div>
+   </div>
+   <div class="card">
+    <div class="card-header">
+      <div class="card-title">
+	<div class="title">Privacy</div>
+      </div>
+    </div>
+    <div class="card-body">
+    <input type="submit" value="Delete Account" class="ui submit button", id="delete-account-btn">
+   </div>
+   </div>
 </div>
 </div>
 {% endblock %}
 {% block scripts %}
 <script>
   $(function () {
-    $('#formerror').modal('toggle')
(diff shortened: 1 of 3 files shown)


agramfort (approved):
LGTM

+1 for MRG when CIs are green

PR #529, opened 2021-05-24
ENH make asynchronous email delivery

### LINKED ISSUE #526: Issue with mail server
The server can hang when sending an email to the participant.
With @agramfort, we think that it could come from the fact that to send an email, we connect to the mail server and send the email. However, this procedure is blocking the rest of the execution. We should migrate to asynchronous processing.

Some details with Flask integration are given there: https://stackoverflow.com/questions/40326651/flask-mail-sending-email-asynchronously-based-on-flask-cookiecutter

glemaitre:
closes #526 

Send email asynchronously.
I assume that tests will fail because we are now sending emails in different Python threads and they will not be registered in the main thread.

I am not sure how we can test it.

### DIFF
--- a/ramp-frontend/ramp_frontend/utils.py
+++ b/ramp-frontend/ramp_frontend/utils.py
@@ -3,12 +3,24 @@ The :mod:`ramp_frontend.utils` provides utilities to ease sending email.
 """
 
 import logging
+from functools import wraps
+from threading import Thread
 
+from flask import current_app
 from flask_mail import Message
 
 from ramp_frontend import mail
 
 logger = logging.getLogger('RAMP-FRONTEND')
+
+
+def async_task(f):
+    """ Takes a function and runs it in a thread """
+    @wraps(f)
+    def _decorated(*args, **kwargs):
+        thr = Thread(target=f, args=args, kwargs=kwargs)
+        thr.start()
+    return _decorated
 
 
 def body_formatter_user(user):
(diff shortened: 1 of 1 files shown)


glemaitre:
It might be possible to have to use `pytest-mock`

```python
mock_start = mocker.patch.object(threading.Thread, 'start', autospec=True)
...
mock_start.called
```

However, it will not fulfill some of the test cases.

ramp-frontend/ramp_frontend/utils.py
```diff
@@ -55,10 +67,24 @@ def send_mail(to, subject, body):
     body : str
         The body of the email.
     """
+    app = current_app._get_current_object()
     try:
```
glemaitre, line 71: I am unsure that having a `try/except` in to make async operation is a good idea.
@rth do you know if it is a blocking operation.

rth: Yeah, it's not. So here you would just exit the clause and the error will never be handled. I think you would need to also add a try: except clause inside _send_async_email. to log a possible error. logging should be thread safe https://docs.python.org/3/howto/logging-cookbook.html#logging-from-multiple-threads

rth:
Closing now that https://github.com/paris-saclay-cds/ramp-board/pull/557 was merged with similar functionality. But don't hesitate to re-open if needed.

PR #528, opened 2021-05-24
WIP Fix circleCI env setup

lucyleeow:

Issue #525, opened 2021-05-24
Documentation is broken


glemaitre:
It seems that CircleCI is broken due to the install environment maybe?

PR #531, opened 2021-05-25
Fix documentation CI

rth:
Closes https://github.com/paris-saclay-cds/ramp-board/issues/525
Supersedes and closes https://github.com/paris-saclay-cds/ramp-board/pull/528


lucyleeow:
I'm not sure the Python version is the problem - it builds fine locally. Was going to check if moving packages to install via conda instead of pip would help.

rth:
I think it might be also the [sphinx 4.0 release](https://pypi.org/project/Sphinx/) that was made last week breaking things.

PR #532, opened 2021-05-25
MAINT Drop Python 3.6 support

rth:
ramp-board is an application so there are less reasons to support 4 versions of Python in CI. ramp.studio uses python 3.7.


agramfort (approved):

agramfort:
thx @rth

PR #535, opened 2021-06-11
Typing fixes + mypy config

rth:
A few fixes so that type checks with mypy could be run without errors.

Also adds the mypy config to pre-commit (use is optional)


rth:
Any objections to this @agramfort @kegl ?

agramfort (approved):

PR #537, opened 2021-06-17
Add User.update_timestamp field + order users to approve

rth:
This makes two changes.
 - adds a `User.update_timestamp` field that is updated for each change of the User table. This is useful mostly for audits. For instance if an admin approved a user by mistake (which happened in the past), it's very difficult to determine which user was changed without this field. This requires a DB migration, but I can do it for ramp.studio

 - Orders the list of users to approve in the admin page by newest at the top.

### DIFF
--- a/doc/whats_new/v0.9.rst
+++ b/doc/whats_new/v0.9.rst
@@ -10,7 +10,7 @@ Changelog
 `ramp-database`
 ...............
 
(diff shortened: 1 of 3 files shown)


frcaud:
Nice improvements @rth. LGTM

frcaud (commented):
code LGTM

Issue #523, opened 2021-05-14
Auto-format code with black 


rth:
It might be a good idea to apply [black](https://pypi.org/project/black/) on the code base, to think less about formatting. This can  also be included in the pre-commit setup to auto-format on each commit.

There aren't many open PR where this would be an issue.

flake8 config would then need to be [adjusted slightly](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#flake8), and the check would need to be added to CI.

agramfort:
No strong feeling

kegl:
I was fine with it till I saw it will replace all single quotes with double. If that can be configured, I'm fine with it.

rth:
> I was fine with it till I saw it will replace all single quotes with double. If that can be configured, I'm fine with it.

The idea of black is have mostly 0 configuration, so that users think less about formatting. So double vs simple quotes cannot be configured.

rth:
An example of how code would look is in #540 . @kegl are single quotes still a blocker?

PR #540, opened 2021-09-03
Apply black to auto-format code

rth:
Closes https://github.com/paris-saclay-cds/ramp-board/issues/523

I still have a few PR planned, and after using auto-formatting tools for some time, doing manual code formatting to keep flake8 happy feels like a waste of time.

I think there are relatively few open PRs, so updating those to use black wouldn't be an issue. 

This uses default black options, because the only way to store config is in a pyproject.toml and it would be a bit strange to add it since the top level folder is not a Python package. But possible to add it if it's a blocker.

Files additionally updated manually 
 - `.flake8` : to be compatible with black
 - `.github/workflows/main.yml`: to check formatting in CI
 - `.pre-commit-config.yaml`: to be able to auto-apply it on each commit
 - `conftest.py`: removed a few unused imports to avoid flake8 errors in pre-commit

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+max-line-length = 88
+extend-ignore = E203
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -124,7 +124,10 @@ jobs:
           python-version: '3.7'
 
       - name: Install dependencies
-        run: pip install flake8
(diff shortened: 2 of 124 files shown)


agramfort (approved):

rth:
@kegl any objections to merging this?

rth:
I'll take the responsibility and merge this to move forward with other planned PRs.

Issue #505, opened 2021-01-11
Using teams in RAMP


rth:
This is an issue to start the discussion on what would be necessary to use teams in RAMP, following a discussion with @kegs.
Having teams would be also quite useful for teaching, where students are often working in pairs.

## Current situation

A short overview, I might have missed something, if so please comment.

Users sign up and are approved individually. They get allocated to teams with just them automatically. Then this 1 person team, requests to join an event, makes submissions and is displayed in the leaderboard.

Generally teams support is partially implemented, and because so far 1 person team is more or less the same as a user, team and users can sometimes be used interchangeably across the API.

DB wise, we have following tables,
 - users : general user information
 - teams : defines the team name and the admin user
 - event_teams : defines team / event membership.

I couldn't find a table mapping users to teams. There is something called `initiator_id`, `acceptor_id` in the `teams` table but that would be enough for teams of 2 at most.

## Option 1

If we wanted to make it work with minimal changes. We could,
1. Add UI to create groups (with approval from all members) before joining events
2. When signing up give the choice to sign up as a 1 person teams or as any other team one is part of.

### Drawbacks

 - The main limitation I see with this approach is that we loose all information about who make which submissions within a group. That's not great particularly for teaching.
 - In terms of UI having to setup teams before having access to the event description is maybe not great for people discovering RAMP.
 - If one changes teams in the middle of the event, all earlier submissions will still be shown as made by a 1 person team.
 - Legally it's also not OK, because since we don't know who contributed what code in a team, we won't be able to remove it if a team member deletes their account. Unless we delete all submissions by a team, which not very user friendly.

### Backward compatibility

 - This would likely require minimal changes in the DB, possibly just adding a user_teams correspondence table would be enough.

## Option 2

That I find conceptually easier, but that might require some (possibly repetitive) changes: all actions on RAMP are made by users not teams, including even sign-up, submissions etc.

Once in the event, users can create a team that's valid only for the current event. It impacts only how submissions are labeled on the leaderboard. There is an option to switch between individual and team view on the leaderboard. If a user deletes their account their submission are removed, which might impact the best score of the team they were in.

### Drawbacks

 - code replacements to use users not teams everywhere
 - possibly require a bit more work from admins to validate `N` users than `N/<team_size>` teams.
 - DB migration (cf below)

### Backward compatibility
 
[...]

PR #545, opened 2021-10-13
Add teams /1: database definitions 

rth:
This is part 1 of the effort to add teams discussed in https://github.com/paris-saclay-cds/ramp-board/issues/505#issuecomment-853725196 , and is a subset of changes [from this PR](https://github.com/symerio/ramp-board/pull/2)

For now this updates changes the DB definitions, to make the review easier,
 - Removes the previous team link mechanism from the `Team` table (`acceptor_id`, `initiator_id`) which wasn't used
 - Add `Team.is_individual` field to distinguish between individual and non individual teams
 - Add a new `UserTeam` table which does many to many relationships between teams and users.

This would require a DB migration, but there is already a commit on main which would require it.

### DIFF
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -1,0 +1,2 @@
+pycodestyle:
+    max-line-length: 88 # black defaults


agramfort:
thx @rth

PR #547, opened 2021-11-06
Add teams 2/3: Add functions to manage teams

rth:
This a follow up on  https://github.com/paris-saclay-cds/ramp-board/pull/545 and adds ramp_database functions to create/join/leave and invite to teams.

Toward https://github.com/paris-saclay-cds/ramp-board/issues/505 

Again it should be backward compatible, as long as those functions are not called (which will be done in a third PR).

TODO: 
 - [ ]  add tests for the few lines not included in test coverage.

### DIFF
--- a/ramp-database/ramp_database/model/event.py
+++ b/ramp-database/ramp_database/model/event.py
@@ -210,8 +210,11 @@ class Event(Model):
         each team."""
         self.n_submissions = 0
         for event_team in self.event_teams:
-            # substract one for starting kit
-            self.n_submissions += len(event_team.submissions) - 1
+            if event_team.team.is_individual:
+                # substract one for starting kit
+                self.n_submissions += len(event_team.submissions) - 1
+            else:
+                self.n_submissions += len(event_team.submissions)
 
     @property
     def Predictions(self):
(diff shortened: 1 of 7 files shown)


BenjaminHabert:
Hello @rth. I had a quick look at these diffs as I said over email. Sorry if these inputs are not relevent : I am not very familiar with the code base.

I am under the impression that the current implementation is slighly different than what you described in https://github.com/paris-saclay-cds/ramp-board/issues/505#issuecomment-853725196 

The `test_team::test_add_signup_leave_team()`  shows the following workflow:
- user creates a team
- team is signed-up to an event

However the initial plan was that a team is only created in the context of a given event. Did this change ? 

If the initial plan is still what you are aiming for : 
-  I think the method `tools/team.py::add_team()` should be renamed to `add_team_to_event()` and create at the same time the `Team` and `TeamEvent` objects. It should not be used to created is_individual teams (as these are created when a user is created). It could also check that the user making this request is allready registered in this event.
- as a consequence`sign_up_team()` should not require the optional `user_name` param as this would only be used
  for is_individual teams.

Another note : I think you might still end up with a user belonging to several multi-user teams in a single event:
- there is a mechanism in `tools/teams.py::respond_team_invite()` to leave the previously joined team.
- however I think a user can create multiple teams in a single event (and be the admin of each of those).
  - this can be usefull if a teacher wants to create all the teams
  - however it will be difficult for this user to make a submission: which team are they submitting for ? (`select_event_team_by_user_name` returns the first team only).

Let me know if I can help

rth:
Thanks a lot for the review @BenjaminHabert, will try to respond to it soon.

PR #552, opened 2021-12-07
Fix numerical issue when using one fold

AGrigis:
Numerical issue when using one fold in 'get_cv'.

Everything happens in the ramp_database.tools.leaderboard module in the _compute_leaderboard function.
When there is only one fold in a submission the data looks like this:

```python
>>> print(df_scores)
          rmse
fold step        
0     train  24.61
       valid  25.83
       test   25.12
```

Then the code makes statistics by step (train,valid,test):

```python
>>> print(df_scores.groupby("step").describe())
       rmse                                              
           count   mean std    min    25%    50%    75%    max
step                                                     
test    1.0  25.12 NaN  25.12  25.12  25.12  25.12  25.12
train   1.0  24.61 NaN  24.61  24.61  24.61  24.61  24.61
valid   1.0  25.83 NaN  25.83  25.83  25.83  25.83  25.83
```

Now the troubles start (std=Nan).
When the .stack() method is used,the default options is .stack(dropna=True).
This point removes the std data from the DataFrame which causes the following error:

```
Traceback (most recent call last):
  File "/ramp_deployment/test.py", line 173, in <module>
    df = df[col_ordered]
  File "/root/miniconda3/lib/python3.9/site-packages/pandas/core/frame.py", line 3464, in __getitem__
    indexer = self.loc._get_listlike_indexer(key, axis=1)[1]
  File "/root/miniconda3/lib/python3.9/site-packages/pandas/core/indexing.py", line 1314, in _get_listlike_indexer
    self._validate_read_indexer(keyarr, indexer, axis)
  File "/root/miniconda3/lib/python3.9/site-packages/pandas/core/indexing.py", line 1377, in _validate_read_indexer
    raise KeyError(f"{not_found} not in index")
KeyError: "['std public rmse', 'std private rmse'] not in index"
```

To solve this issue you can:

* replace NaN by zeros (the one currently implemented in the PR):

```python
df_scores_std = df_scores.groupby("step").std()
df_scores_std.fillna(0, inplace=True)
```

* keep the NaN in the DataFrame:

```python
df_scores_std = (
            df_scores_std.loc[list(map_renaming.keys())]
            .rename(index=map_renaming)
            .stack(dropna=False)
            .to_frame()
            .T)
```
This choice must surely be conditioned with the frontend (are NaNs managed?).


agramfort (approved):
can you just add a line to change log in https://github.com/paris-saclay-cds/ramp-board/blob/master/doc/whats_new/v0.9.rst ?

🙏 @AGrigis

agramfort (approved):
@rth @kegl @albertcthomas ok for you?

albertcthomas:
What about using `df_scores_std = df_scores.groupby("step").std(ddof=0)` so that you use `1/n` instead of `1/n-1` (`numpy.std` uses `ddof=0` by default I think).

kegl:
I'm fine with either the current change and the suggestion of @albertcthomas

agramfort:
both are fine with me. The solution here is at least not an API change but really a bug fix

albertcthomas:
I have also no strong opinion. Maybe keeping the NaNs in the dataframe could avoid silent failures.

agramfort:
if we keep the NaN is breaks on a ramp where there is only one fold

>

albertcthomas:
> To solve this issue you can:
> 
>     * replace NaN by zeros (the one currently implemented in the PR):
> 
> 
> ```python
> df_scores_std = df_scores.groupby("step").std()
> df_scores_std.fillna(0, inplace=True)
> ```
> 
>     * keep the NaN in the DataFrame:
> 
> 
> ```python
> df_scores_std = (
>             df_scores_std.loc[list(map_renaming.keys())]
>             .rename(index=map_renaming)
>             .stack(dropna=False)
>             .to_frame()
>             .T)
> ```

what I meant is that between these two solutions the first one replaces NaNs by 0, whereas the second one keeps the NaNs. The first solution could make it harder to spot an issue in the computed results: for instance NaNs could propagate and at the end they are replaced by 0s.

agramfort:
whatever works. Do want you think is best. I trust your judgement :)

>

albertcthomas:
Maybe @AGrigis has a preference :)

AGrigis:
Thanks you for your feedback.
I don't have a strong experience with the ramp framework.
I let you choose the best option to preserve the coherence of the different plugged ramp modules.

agramfort:
@AGrigis can you fix the conflict in what's new?

then let's move on and merge as is. 

thx

rth:
CI if failing because I think the Openml server seem to have issues, I'll retry later. Otherwise, I moved the changelog entry so this should be ready for merging.

agramfort:
I restarted the CI. 

@rth we'll need this PR for the next release

rth:
CI is green, merging, thanks!

PR #553, opened 2021-12-11
new tag for google analytics

agramfort:
we need to update GA JS code


glemaitre:
If we want to comply with the CNIL, we should remove google analytics, isn't it?
xref: https://www.cnil.fr/en/use-google-analytics-and-data-transfers-united-states-cnil-orders-website-manageroperator-comply

agramfort:
sklearn uses GA but with limitation.

I would not consider an alternative and to me it's useful to have it

rth:
Thanks, merging.

PR #556, opened 2022-01-11
first try with async email

agramfort:
the pain was too much....


glemaitre:
closes in favor of #557

PR #557, opened 2022-01-14
ENH use threadpool to send emails in the background

agramfort:
joint work with @tomMoral 

@rth @glemaitre thoughts?


rth:
Thanks for this! I also think that a threadpool with a email queue is the best way to go. It's probably it's better than creating a thread at each call as in https://github.com/paris-saclay-cds/ramp-board/pull/529 since at least it gives a single global instance with tasks. 

The only thing is that tests would need to be adjusted to wait for all tasks to be complete in the pool https://stackoverflow.com/a/21143809/1791279 before making assertions.

ramp-frontend/ramp_frontend/utils.py
```diff
@@ -67,6 +68,6 @@ def send_mail(to, subject, body):
         msg = Message(subject)
         msg.body = body
         msg.add_recipient(to)
-        mail.send(msg)
+        current_app.pool.submit(copy_current_request_context(mail.send), msg)
     except Exception as e:
         logger.error("Mailing error: {}".format(e))
```
tomMoral, line 73: Good practice to make sure we have both logs and we are reliable in sending email/not crashing the server:

- we should add a `ensure_running` function that restart the pool if it is not running.
- we should add a way to log exceptions in send mail, maybe simply by putting the try except in the threadpool

```suggestion
        def send(msg):
            try:
                mail.send(msg)
            except Exception as e:
                logger.error("Mailing error: {}".format(e))

    current_app.pool.submit(copy_current_request_context(mail.send), msg)
```

ramp-frontend/ramp_frontend/__init__.py
```diff
@@ -58,6 +59,7 @@ def create_app(config):
 
     app = Flask("ramp-frontend", root_path=HERE)
     app.config.update(config)
+    app.pool = ThreadPoolExecutor(max_workers=app.config["THREADPOOL_MAX_WORKERS"])
```
glemaitre, line 62: I added a configuration entry to set the number of workers.

ramp-frontend/ramp_frontend/tests/test_auth.py
```diff
@@ -451,6 +450,13 @@ def test_reset_password(client_session):
                 reset_password_link.find("/reset") :
             ]
 
+    rv = client.post("/reset_password", data={"email": user.email})
```
glemaitre, line 453: This would have sent an email and we would have 2 emails in the outbox instead of one. I assume that the test was passing because we were blocking the execution in some way :)

ramp-frontend/ramp_frontend/tests/test_ramp.py
```diff
@@ -472,6 +474,8 @@ def test_ask_for_event_mail(client_session):
                 rv = client.post("problems/iris/ask_for_event", data=data)
                 assert rv.status_code == 302
                 # check that the email has been sent
+                # shutdown the threadpool and wait for the future (email) to be sent
+                client.application.pool.shutdown(wait=True)
```
glemaitre, line 478: Since we shutdown the pool, we make as well sure that we restart it properly.

ramp-frontend/ramp_frontend/utils.py
```diff
@@ -3,14 +3,36 @@
...
+    current_app,
+)
 from flask_mail import Message
 
 from ramp_frontend import mail
 
 logger = logging.getLogger("RAMP-FRONTEND")
 
 
+def ensure_threadpoolexecutor_is_running(pool):
```
glemaitre, line 19: @tomMoral could you provide some feedback here because I thought it was what you meant but I never really used the pool executor before.

ramp-frontend/ramp_frontend/tests/test_utils.py
```diff
@@ -46,18 +48,6 @@ def client_session(database_connection):
         Model.metadata.drop_all(db)
 
 
-@_fail_no_smtp_server
```
glemaitre, line 49: This test cannot work anymore without having a view and some requests.
However, we can rely on the integration test that makes a similar check for each email sent.

glemaitre:
Merging and I will prepare a minor release

ramp-frontend/ramp_frontend/utils.py
tomMoral: This seems fine like this.

tomMoral (commented):
Just a question on where the `app.pool` object but other wise looks quite good

ramp-frontend/ramp_frontend/utils.py
```diff
@@ -63,10 +85,17 @@ def send_mail(to, subject, body):
...
+
+    @copy_current_request_context
+    def send(msg):
+        try:
+            mail.send(msg)
+        except Exception as e:
+            logger.error(f"Mailing error: {e}")
+        return "Successfully sent email"
+
+    current_app.pool = ensure_threadpoolexecutor_is_running(current_app.pool)
```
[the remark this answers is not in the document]
glemaitre: it is created when instantiated the app in `__init__.py`. Launching the frontend will always create this attribute. Then, what could happen is that the pool was shut down for some reason.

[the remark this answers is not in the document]
tomMoral: Ok sorry, bad ctrl+F on my side.

Perfect then :)

PR #564, opened 2022-02-21
add delete_problem command to ramp-database

frcaud:
- add a delete_problem command to CLI of ramp-database
- infos corrections

### DIFF
--- a/ramp-database/ramp_database/cli.py
+++ b/ramp-database/ramp_database/cli.py
@@ -303,11 +303,25 @@ def add_submission(config, event, team, submission, path):
     show_default=True,
     help="Configuration file YAML format containing the database information",
 )
+@click.option("--problem", help="Name of the problem")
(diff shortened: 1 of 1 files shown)


agramfort:
Diff looks good. Can you just add a test and a what’s new entry ? 🙏

ramp-database/ramp_database/tests/test_cli.py
```diff
@@ -198,6 +198,40 @@ def test_add_event(make_toy_db):
...
+        main,
+        [
+            "delete-problem",
+            "--config",
+            database_config_template(),
+            "--problem",
+            "iris",
+        ],
+        catch_exceptions=False,
+    )
```
agramfort, line 231: any chance you can assess in the database that the problem is available after you add it and then gone once you remove it? see how it's done elsewhere.

PR #565, opened 2022-02-25
DOC improve EC2 pipeline using mamba

tomMoral:
A pass on the EC2 image builder to remove some unneeded weight and use mamba to accelerate image building,


frcaud:
Thanks @tomMoral !

tomMoral:
I don't have the push right here so not sure who will merge :)

PR #571, opened 2022-03-27
FIX local workers with niceness 5

tomMoral:
Having worker that are lower priority will make running and managing the server smoother.


rth (approved):
Sounds good. I don't know if anyone tried to run ramp-board on windows, in which case we should do this conditionally. 

Otherwise, CI should be fixed in https://github.com/paris-saclay-cds/ramp-board/pull/572

agramfort (approved):

Issue #185, opened 2019-03-08
Define minimum versions of dependencies


jorisvandenbossche:
Point of feedback from the workshop: for quite some packages, we require a recent version, but we don't specify this. So people using their installed packages instead of a fresh environment ran into some errors.
(eg click for the dash vs underscore? `sqlachemy` needs `model_class` keyword. `pandas.concat` needs `sort` keyword)

Ideally we should define and test minimum versions.

rth:
> Ideally we should define and test minimum versions.

Or maybe just pin dependencies? This is an application that should ideally be installed in a separate well-controlled environment, not a library that could be installed everywhere.

Otherwise, with current unpinned versions of dependencies (and the number of dependencies), things are likely to break in 6 month or so, unless someone maintains it continuously.

jorisvandenbossche:
Yep, that is actually true. And certainly given that we can train submissions in a different environment (where you want to have the latest), we can properly pin for ramp-board itself.

PR #572, opened 2022-04-01
Fix CI

agramfort:
testings CIs from master


rth:
I pinned most requirements since this is an application not a library (i.e. people are not supposed to import ramp-board in their code), so we can fully control dependencies in a separate env. Otherwise, it will keep breaking. https://github.com/paris-saclay-cds/ramp-board/issues/185#issuecomment-477772582

Waiting for the CI to finish but it should be OK I think.

agramfort:
thx @rth

PR #574, opened 2022-04-20
[DOC] adding note for ami_image_name generic name

frcaud:
When filling 'config.yml' file for an event and when using AWS worker, one can put generic AMI name under 'ami_image_name'. It will automatically point to the latest version of the image created running the pipeline.


agramfort (approved):

agramfort:
@frcaud any idea why CIs fail here?

tomMoral (approved):
LGTM!

frcaud:
@agramfort green now. Seems GitHub actions were having issues previously.

agramfort:
Thx @frcaud

PR #576, opened 2022-05-21
FIX add hyperlink to log when submission failed

### LINKED ISSUE #575: Missing error traceback when a submission failed
In the past, we were providing the traceback with the log when a submission was failing.
Someone could investigate the issue:

image

Now, there are no hyperlinks with the `training_error`.
@rth @agramfort do you know if we change this?

glemaitre:
closes #575

I suspect this is the source of the error that was introduced in https://github.com/paris-saclay-cds/ramp-board/pull/366

We should probably add a non-regression test that checks that we have a link to a hyperlink with the `failed` leaderboard.

### DIFF
--- a/ramp-database/ramp_database/tools/tests/test_leaderboard.py
+++ b/ramp-database/ramp_database/tools/tests/test_leaderboard.py
@@ -172,7 +172,7 @@ def test_get_leaderboard(session_toy_db):
     )
     assert leaderboard_failed.count("<tr>") == 1
     # check that we have a link to the log of the failed submission
-    assert re.match(r"<a href=/.*/error.txt>", leaderboard_failed)
+    assert re.match(r".*<a href=/.*/error.txt>.*", leaderboard_failed, flags=re.DOTALL)
 
     # the remaining submission should be successful
     leaderboard_public = get_leaderboard(session_toy_db, "public", "iris_test")
--- a/ramp-database/ramp_database/tools/tests/test_leaderboard.py
+++ b/ramp-database/ramp_database/tools/tests/test_leaderboard.py
@@ -170,6 +170,10 @@ def test_get_leaderboard(session_toy_db):
         session_toy_db, "failed", "iris_test", "test_user"
     )
     assert leaderboard_failed.count("<tr>") == 1
+    # check that we have a link to the log of the failed submission
+    print(leaderboard_failed)
+    assert False
+    assert "<a href=" in leaderboard_failed
 
     # the remaining submission should be successful
     leaderboard_public = get_leaderboard(session_toy_db, "public", "iris_test")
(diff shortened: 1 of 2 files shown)


ramp-database/ramp_database/tools/tests/test_leaderboard.py
```diff
@@ -170,6 +170,10 @@ def test_get_leaderboard(session_toy_db):
         session_toy_db, "failed", "iris_test", "test_user"
     )
     assert leaderboard_failed.count("<tr>") == 1
+    # check that we have a link to the log of the failed submission
+    print(leaderboard_failed)
+    assert False
+    assert "<a href=" in leaderboard_failed
```
glemaitre, line 176: OK so the string is:

```
<a href=/abcb5aeab7e78cd43636a10d6856dcdf4ae886c6/error.txt>
```

So we can make a regex `<a href=/*/error.txt>`

glemaitre, line 176:
```suggestion
    assert re.match("<a href=/.*/error.txt", leaderboard_failed)
```

ramp-database/ramp_database/tools/tests/test_leaderboard.py
```diff
@@ -170,6 +171,8 @@ def test_get_leaderboard(session_toy_db):
         session_toy_db, "failed", "iris_test", "test_user"
     )
     assert leaderboard_failed.count("<tr>") == 1
+    # check that we have a link to the log of the failed submission
+    assert re.match("<a href=/.*/error.txt", leaderboard_failed)
```
glemaitre, line 175:
```suggestion
    assert re.match(r"<a href=/.*/error.txt>", leaderboard_failed)
```

ramp-database/ramp_database/tools/tests/test_leaderboard.py
```diff
@@ -170,6 +171,8 @@ def test_get_leaderboard(session_toy_db):
         session_toy_db, "failed", "iris_test", "test_user"
     )
     assert leaderboard_failed.count("<tr>") == 1
+    # check that we have a link to the log of the failed submission
+    assert re.match(r"<a href=/.*/error.txt>", leaderboard_failed)
```
glemaitre, line 175:
```suggestion
    assert re.match(r".*<a href=/.*/error.txt>.*", leaderboard_failed, flags=re.DOTALL)
```

agramfort:
Thx @glemaitre 

I suggest we make a minor release and update the server ok?

glemaitre:
> I suggest we make a minor release and update the server ok?

I would like to tackle this one before that:

https://github.com/paris-saclay-cds/ramp-board/issues/577

rth:
@glemaitre @frcaud  I'll tag the 0.10.1 release with this fix and deploy this if you don't mind. I haven't realized it wasn't deployed, and it's a regression.

cc @jjerphan

glemaitre:
Fine with me.

frcaud:
LGTM !

PR #578, opened 2022-06-17
up size limit for data in submission_file_types

frcaud:
Up the size limit for data in submission_file_types, mainly to upload bigger weights files for NN (from 100M to 1G)

### DIFF
--- a/ramp-database/ramp_database/testing.py
+++ b/ramp-database/ramp_database/testing.py
@@ -218,7 +218,7 @@ def setup_files_extension_type(session):
     submission_file_types = [
         ("code", True, 10**5),
         ("text", True, 10**5),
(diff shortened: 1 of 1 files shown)


rth:
I'm not sure it would have any effect for us though, as this is only used when initiallizing the DB, I think? For an existing DB we need to change the value in the DB directly.

frcaud:
I also changed value of max_size in DB directly and it solves our pb

agramfort:
@frcaud tests complain

frcaud:
CIs are green !

agramfort:
I am surprised you have to modify things just here. You have references elsewhere to the 100MB limit eg in test and maybe in the doc ?

PR #579, opened 2022-06-17
remove json files in pkgs/cache as a workaround for permission error

frcaud:
Workaround for a mamba permission error with json files in pkgs/cache


frcaud:
Everyone OK with this change ?

PR #589, opened 2023-09-07
Add 'submission_id' and 'Kill Sub' columns in New submissions leaderboard

frcaud:
Adding a Kill Submission button in New submissions leaderboard to be able to terminate a submission.
First PR is adding a 'submission_id' column and adding a 'Kill Sub' column with clickable buttons on the HTML rendering of the New submissions leaderboard (leaderboard and My submissions pages on frontend)

### DIFF
--- a/ramp-database/ramp_database/tools/leaderboard.py
+++ b/ramp-database/ramp_database/tools/leaderboard.py
@@ -439,7 +439,7 @@ def get_leaderboard(
             columns = [
                 "team",
                 "submission",
-                "submission_id",
+                "id",
                 "submitted at (UTC)",
                 "state",
                 "waiting list",
@@ -448,7 +448,7 @@ def get_leaderboard(
             columns = [
                 "team",
                 "submission",
-                "submission_id",
+                "id",
                 "submitted at (UTC)",
                 "error",
             ]
@@ -487,7 +487,7 @@ def get_leaderboard(
                 return f"<button onclick=\"alert(('You killed ' \
                     + 'submission {cell_value}!'))\">{cell_value}</button>"
 
-            df["Kill Sub"] = df["submission_id"].apply(
+            df["Stop"] = df["id"].apply(
                 lambda x: create_button(x)
             )
     else:
--- a/ramp-database/ramp_database/tools/leaderboard.py
+++ b/ramp-database/ramp_database/tools/leaderboard.py
@@ -17,7 +17,9 @@ from .submission import get_submission_max_ram
 from .submission import get_time
 
 
-width = -1 if version.Version(pd.__version__) < version.Version("1.0.0") else None
+width = (
+    -1 if version.Version(pd.__version__) < version.Version("1.0.0") else None
+)
 pd.set_option("display.max_colwidth", width)
 
 
@@ -47,7 +49,8 @@ def _compute_leaderboard(
     record_score = []
     event = session.query(Event).filter_by(name=event_name).one()
     map_score_precision = {
-        score_type.name: score_type.precision for score_type in event.score_types
+        score_type.name: score_type.precision
+        for score_type in event.score_types
     }
     for sub in submissions:
         # take only max n bag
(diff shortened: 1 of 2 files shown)


rth (commented):
Thanks @frcaud! That's the idea, though we need to figure out why CI fails and fix it.

glemaitre:
My first guess (looking at the last history commit in `main`) would be to check the `pytest`. I would assume that this PR is using the latest `pytest` and they might have issue deprecation warning about the way we used the fixture and we did not change it on time.

glemaitre:
I see that `pytest==6.2.5` does not raise the error for the fixture. So it should be something new.

The `conf.py` file is here: https://github.com/paris-saclay-cds/ramp-board/blob/9b1d738e357a708f0387d63de4d8577216edd565/conftest.py#L11

glemaitre:
It might be due to https://github.com/pytest-dev/pytest/issues/11261

glemaitre:
You need to add an empty `pytest.ini` in the root folder

frcaud:
Thanks @glemaitre !
 Frontend error is resolved. 
Still some issue on test database

ramp-database/ramp_database/tools/leaderboard.py
```diff
@@ -427,12 +439,19 @@ def get_leaderboard(
             columns = [
                 "team",
                 "submission",
+                "submission_id",
```
rth, line 442: Maybe let's just call it `id` so it would be a less wide column in the resulting html table? It's shown under a section of New submission, so there is no need to repeat "submission" in column names I think.

ramp-database/ramp_database/tools/leaderboard.py
```diff
@@ -461,12 +481,23 @@ def get_leaderboard(
             for sub in submissions
         ]
         df = pd.DataFrame(data, columns=columns)
+        if leaderboard_type == "new":
+
+            def create_button(cell_value):
+                return f"<button onclick=\"alert(('You killed ' \
+                    + 'submission {cell_value}!'))\">{cell_value}</button>"
+
+            df["Kill Sub"] = df["submission_id"].apply(
```
rth, line 490: Again just "Kill" or maybe better "Stop" would be enough. It's in a table of submissions already.

ramp-database/ramp_database/tools/leaderboard.py
```diff
@@ -461,12 +481,23 @@ def get_leaderboard(
...
+            )
     else:
         # make some extra filtering
         submissions = [sub for sub in submissions if sub.is_public_leaderboard]
         if not submissions:
             return None
-        competition_type = "public" if "public" in leaderboard_type else "private"
+        competition_type = (
+            "public" if "public" in leaderboard_type else "private"
+        )
```
rth, line 500: I think if `leaderboard_type not in ['new', 'failed']` we need to delete the extra column "id" we created before `_compute_competition_leaderboard` is run otherwise it will show in all other leaderboards

rth (commented):
Thanks @frcaud ! For the failing test we need to edit `test_get_leaderboard` to add the extra column we added in,
```py
>       assert (
            """<th>team</th>
          <th>submission</th>
          <th>submitted at (UTC)</th>
          <th>error</th>"""
            in leaderboard_failed
        )
```
assertion. The pytest log is a bit confusing, but I think that's actually the only failing test. The other exceptions you see in the logs, are the expected exceptions that are obtained when running a bad submission and can be ignored.

A few more comments below, otherwise LGTM.

ramp-database/ramp_database/tools/leaderboard.py
frcaud: Since we create this extra column in ```elif leaderboard_type in ["new", "failed"]:``` it seems that we don't need to delete it in other leaderboard types

frcaud:
Thanks @rth ! Tests OK. Seems there is one more error with codecov.

frcaud:
CIs are green on rerun. Thanks @glemaitre and @rth !

rth:
Thanks! Linting CI is still failing though. Can you run black on it?

rth (approved):

PR #590, opened 2023-12-06
Allow using mean CV score as main score in the leaderboard

rth:
This allows to optionally use mean CV score as main score, instead of bagged score, in the leaderboard by setting,
```
leaderboard_score_kind = "mean"
```
in the kit `problem.py`.

The use case is in particular kits that are not able to compute bagged scores such as [stroke_lesions](https://github.com/ramp-kits/stroke_lesions)

cc @frcaud


frcaud:
Great solution ! Thanks @rth !

PR #591, opened 2024-02-29
add a cmd to find error in _training_successful

frcaud:
Add a cmd to find error during check if training has been successful on the EC2 instance -> if error found then it returns ```false```


rth (approved):
Thanks @frcaud . LGTM, but let's merge https://github.com/paris-saclay-cds/ramp-board/pull/592 first and sync here to fix CI.

PR #592, opened 2024-02-29
Fix CI

rth:
- Switch to fetching starting kit repos via HTTP rather than using git clone. As far as I can tell all those git clone in CI tests were triggering some github safety blocking the requests, leading the CI to hang.
 - Fix for leap year on 29 february,
```
    NOW.replace(year=NOW.year + 1),
E   ValueError: day is out of range for month 
```

### DIFF
--- a/ramp-database/ramp_database/testing.py
+++ b/ramp-database/ramp_database/testing.py
@@ -183,7 +183,7 @@ def setup_ramp_kit_ramp_data(
         shutil.rmtree(problem_kit_path, ignore_errors=True)
 
     ramp_kit_url = (
-        f"https://github.com/ramp-kits/{problem_name}" f"/archive/refs/heads/master.zip"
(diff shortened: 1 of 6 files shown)


ramp-database/ramp_database/testing.py
```diff
@@ -172,8 +195,10 @@ def setup_ramp_kit_ramp_data(
                 'it, you need to set "force=True".'
             )
         shutil.rmtree(problem_data_path, ignore_errors=True)
-    ramp_data_url = "https://github.com/ramp-data/{}.git".format(problem_name)
-    Repo.clone_from(ramp_data_url, problem_data_path, **kwargs)
+    ramp_kit_url = (
```
frcaud, line 198: ```ramp_data_url``` here instead of ```ramp_kit_url``` ?

frcaud (commented):
Thanks @rth ! LGTM
Just see the comment line 198 of testing.py

